PR #11, opened 2020-04-27
Replaced all hardcoded configuration options (PORT, AUDIO, WPM) with …

pwdirks:
…uses of 'config' module.


AESilky:
Thanks for updating all of the commands. I didn't want to do that until it was tested on Mac and I don't get my Mac back until mid-week.

PR #24, opened 2020-05-01
Change string formatting of `f` to String.format() to remove requireme…

AESilky:
…nt of Python 3.6


AESilky:
This is a straightforward change and I ran `configure` to list and set values and reviewed the output for correctness, so I am merging it directly.

PR #29, opened 2020-05-03
Pwd 16 clock announce timing

pwdirks:
This change restructures the Clock.py program to be able to announce on arbitrary intervals and in more natural formats:

   "midnight"
   "Nine o"clock"
   "A quarter past nine"
   "9:20"
   "Half past nine"
   "A quarter to ten"
   "noon"

In the process I've created an experiment in the command line parsing, changing 'config.py' to create overrides for the options defined in the config file and changing Clock.py to builds its command line parser with the defaults as they're set in the config file but with the option to override them.

In addition I added an 'argsparser'-based command line parser for Clock-specific options:
    -b \<begin time\>
    -e \<end time\>
    -i \<interval\>

The idea being you can start the clock with

    python3 Clock.py -s 20 -i 15 -a OFF

to start it running at 20 WPM (regardless of configured defaults), at 15 minute intervals, with the audio always OFF, on the default configured port.

I'd welcome all comments, especially on choices of variable and global names.  Existing code had some all-caps variables, and ideal choice of case for globals in 'config' is unclear to me.

### DIFF
--- a/Clock.py
+++ b/Clock.py
@@ -130,7 +130,9 @@ try:
     
     port = args.Port # serial port for KOB interface
     speed = args.Speed  # code speed (words per minute)
+    print("args.Sound is", args.Sound)
     sound = True if args.Sound == 'ON' else False
+    print("sound =", sound)
     start_time = hms_to_seconds(int(args.Begin/100), args.Begin % 100, 0)  # start time (sec)
     end_time = hms_to_seconds(int(args.End/100), args.End % 100, 0)  # end time (sec)
     annc_interval = args.Interval * 60      # announcement interval (sec)
--- a/Clock.py
+++ b/Clock.py
@@ -110,7 +110,7 @@ def announcement(hours, minutes):
 # Send a message as morse on the sounder:
 #
 def announce(s, kob, sender):
-    print("> ", s)
+    print(">", s)
     for c in s:
         code = sender.encode(c)
         kob.sounder(code)
@@ -150,16 +150,16 @@ try:
         next_time = round_up(hms_to_seconds(t.tm_hour, t.tm_min, t.tm_sec), annc_interval)  # next announcement (sec)
         if next_time < start_time:
             next_time = start_time
-        print("Next time announcement at {0}:{1:02d}".format(clock_hour(next_time), clock_minutes(next_time)))
+        # print("Next time announcement at {0}:{1:02d}".format(clock_hour(next_time), clock_minutes(next_time)))
         now = hms_to_seconds(t.tm_hour, t.tm_min, t.tm_sec)  # current time (sec)
         if next_time <= end_time:
             if now < next_time:
-                print("sleeping for {0} seconds until next time announcement...".format(next_time - now))
+                # print("sleeping for {0} seconds until next time announcement...".format(next_time - now))
                 time.sleep(next_time - now)
             msg = announcement(clock_hour(next_time), clock_minutes(next_time))
             announce(msg, myKOB, mySender)
         else:
-            print("sleeping for {0} seconds until midnight...")
+            # print("sleeping for {0} seconds until midnight...")
             time.sleep(24*3600 - now)  # wait until midnight and start over
 except KeyboardInterrupt:
     print()
--- a/Clock.py
+++ b/Clock.py
@@ -33,58 +33,134 @@ Serial port, code speed, and audio preferences should be specified by running th
 'configure.sh' script or executing 'python3 Configure.py'.
 """
 
+NUMBER = [ "oh", "one", "two", "three", "four", "five", "six", "seven", "eight", "nine", "ten", "eleven", "twelve"]
+
+#
+# Convert hours/minutes/seconds since midnight to seconds since midnight:
+#
+def hms_to_seconds(hours, minutes, seconds):
+    return hours * 3600 + minutes * 60 + seconds
+
+#
+# Convert from numeric representation to text ('9' to 'nine')
+#
+def number_to_text(n):
+    if (n < 0) or ( n > 12):
+        return ""
+    return NUMBER[n]
+#
+# Convert 0-24 hours to 0-12 hours by subtracting 12 if needed:
+#
+def truncate_hours(h):
+    if h <= 12:
+        return h        # 0-12 is returned as-is
+    else:
+        return h % 12   # 13-on is truncated to 1-on
+
+#
+# Convert seconds since midnight to corresponding clock hour
+#
+def clock_hour(t):
+    return int(t / 3600)
+
+#
+# Convert seconds since midnight to the corresponding clock minute
+#
+def clock_minutes(t):
+    return int(int(t % 3600) / 60)
+
+#
+# Take an starting point and round up to the next interval-multiple:
+#
+def round_up(start, interval):
+    return start - (start % interval) + interval # Round up to next interval (sec)
+
+#
+# Generate a proper time announcement:
+#
+#   "midnight"
+#   "Nine o'clock"
+#   "A quarter past nine"
+#   "9:20"
+#   "Half past nine"
+#   "A quarter to ten"
+#   "noon"
+#
+def announcement(hours, minutes):
+    msg = "The time is "
+    if minutes == 0:
+        if hours == 0:
+            msg += "midnight"
+        elif hours == 12:
+            msg += "noon" + 12 * "L "
+        else:
+            msg += number_to_text(truncate_hours(hours)) + " oclock     " + hours * "L "
+    elif minutes == 15:
+        msg += "a quarter past " + number_to_text(truncate_hours(hours))
+    elif minutes == 30:
+        msg += "half past " + number_to_text(truncate_hours(hours))
+    elif minutes == 45:
+        msg += "a quarter to " + numer_to_text(truncate_hours(hours + 1))
+    else:
+        # Last resort: just generate the time in hh:mm format
+        msg += "{hour}:{minute:02d}".format(hour=hours, minute=minutes)
+    return msg
+
+#
+# Send a message as morse on the sounder:
+#
 def announce(s, kob, sender):
+    print("> ", s)
     for c in s:
         code = sender.encode(c)
         kob.sounder(code)
 
 try:
+    import argparse
     import sys
     import time
     from pykob import config, kob, morse, log
 
     log.log('Starting Clock')
 
+    clock_parser = argparse.ArgumentParser(parents=[config.WPM_OVERRIDE])
+    clock_parser.add_argument('-b', '--begin', default=900, type=int, help='Beginning of time announcements ', metavar='time', dest='Begin')
+    clock_parser.add_argument('-e', '--end', default=2230, type=int, help='End of time announcements ', metavar='time', dest='End')
+    clock_parser.add_argument('-i', '--interval', default=60, type=int, help='The time announcement interval in minutes', metavar='minutes', dest='Interval')
+    args = clock_parser.parse_args()
+    
     PORT = config.Port # serial port for KOB interface
-    WPM = config.Speed  # code speed (words per minute)
+    WPM = args.Speed  # code speed (words per minute)
     SOUND = config.Sound # whether to enable computer sound for sounder
-    MSGS = [
-        ( 900, 'The time is nine oclock     L  L  L  L  L  L  L  L  L'),
-        (1000, 'The time is ten oclock     L  L  L  L  L  L  L  L  L  L'),
-        (1100, 'The time is eleven oclock     L  L  L  L  L  L  L  L  L  L  L'),
-        (1200, 'The time is twelve oclock     L  L  L  L  L  L  L  L  L  L  L  L'),
-        (1300, 'The time is one oclock     L'),
-        (1400, 'The time is two oclock     L  L'),
-        (1500, 'The time is three oclock     L  L  L'),
-        (1600, 'The time is four oclock     L  L  L  L'),
-        (1700, 'The time is five oclock     L  L  L  L  L'),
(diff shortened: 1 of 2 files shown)


pwdirks:
Oh, BTW - the existing code was a bit inconsistent about the 'SOUND' option - it's stored in the config file as a string ('ON' or 'OFF') but it was retrieved and used as a boolean.  I believe it always came out 'True' as a result.  I changed the code to retrieve it as a string and leave it that way but I had to change my code that consumed it to convert it to a boolean before creating the KOB.

I'll file a separate issue to either have all calling code configured similarly or to change the config option to save and restore a boolean.  The current usage makes it awkward to set the default option for the "--sound" override.

leskerr:
This is fantastic! I haven't tried it yet, but I love the concept from a 'learning the code' pedagogical point of view. I haven't thought through all the details yet, but by including the quarter hour text, for example, the user will learn the letter Q through subliminal reinforcement every half hour. The biggest enemy of learning the code through code practice is boredom. This way the user has a potentially useful tool (maybe as a reminder to stretch every 15 minutes while sitting in front of the computer) and at the same time a periodic refresher on code sequences that become ingrained fairly quickly.

Some users may want the option of inserting Farnsworth spacing while they're still learning.

Another useful option might be the ability to specify the text at the beginning of the announcement. It could be as simple as overriding the default 'the time is' with 'the time is now', which would give more practice with the letter W. With some imagination, the user could change the message to 'get up and stretch', or whatever. The idea is to hear a familiar phrase over and over again.

And speaking of options, what about specifying a 'quiet period'? With my original Clock program, it was hardcoded to only announce the time between 9am and 10pm, which is the way my wife wanted it. But a general program should be more flexible.

pwdirks:
> Some users may want the option of inserting Farnsworth spacing while they're still learning.

I believe that will require a change in the 'kob' class, to take 2 speeds (character speed and overall speed)?  When we add config options for both we can change all code to initialize the kob with the specified speeds, triggering Farnsworth spacing if requested.  Of course that'll be something the user can override on the command line as well.

> Another useful option might be the ability to specify the text at the beginning of the announcement. It could be as simple as overriding the default 'the time is' with 'the time is now', which would give more practice with the letter W. With some imagination, the user could change the message to 'get up and stretch', or whatever. The idea is to hear a familiar phrase over and over again.

Agreed - that'd be a nice addition.  Can you file a separate issue for that request?

> And speaking of options, what about specifying a 'quiet period'? With my original Clock program, it was hardcoded to only announce the time between 9am and 10pm, which is the way my wife wanted it. But a general program should be more flexible.

That's actually covered - I noticed the table had entries going from 0900 to 2200 only so I added "-b for begin" (I put the text in angle brackets and I see the comment dropped it - probably some bad XML parse in GitHub), which defaults to 900 and "-e for end" of the active period, which defaults to 2200 or 2230 (can't recall).  Between "-b", "-e", and "-i" you should be able to get the announcements exactly when you want them.

AESilky:
@pwdirks I didn't test it extensively, but for 'SOUND' I thought I used a parsing option that processed values into a boolean and accepted 'ON, TRUE, YES' as 'True' and 'OFF, FALSE, NO' as False. Did you find that to not be working correctly? I know I tested with 'ON' and 'OFF' and I believe it enabled/disabled the sound as I expected. Once I catch up on email I'll test it out again with my current installation.

AESilky:
@pwdirks everything in the config (.ini) file is stored as a string. I believe that is a requirement of the 'configparser' class (isn't it?). But there are access methods that handle conversions.

Please let me know if that isn't the case.

pwdirks:
Hi Ed - I checked the text file and it was indeed t.  IIRC it was coming out as “True” always and my change made it work for me but let me check a bit more carefully.  I didn’t see it reading “OFF” as False.

pwdirks:
@AESilky I have no idea what I thought I was seeing - there seems to be nothing wrong with the way pykob/config.py is setting 'Sound' and it's coming out properly as a boolean.  I've reverted the changes I made to that part of the config parsing logic, leaving only the command line argument overrides to be picked up by clients, which I *DID* fix up to convert the boolean 'Sound' back to an appropriate default ('ON' or 'OFF').  See if this isn't cleaner overall.

With the possibility of an override comes the choice to look at config.Sound (boolean) or args.Sound ('ON' or 'OFF'), defaulting to what's configured.  I've change the Clock.py code to take a fresh look at the args.Sound setting and convert that back to a boolean.  That'll be necessary for any program switching to using command-line overrides but for now programs using 'config.Sound' as a boolean should be just fine.

AESilky:
I agree with @pwdirks on his earlier comment on Farnsworth spacing. That should be one of the 'common' configuration/preference options that all commands will pick up automatically by using the config module. The combination of 'speed' and 'spacing?' would set the Farnsworth value to be used. This would then apply to any program you run (Sample, Clock, News, etc.) and can be overridden on any particular instance/run of the application.

pwdirks:
Question: what would folks expect around special times (noon and midnight)?  Do people say "It's a quarter to noon" or "a quarter to 12"?  "quarter/half past noon" or "quarter/half past 12"?

I'm guessing at least for noon it's the latter - "noon" is only for that exact time point - otherwise it's "quarter to twelve".  But what about midnight?  Nobody says "quarter to zero"?  Or "quarter past zero" - it's got to be "a quarter to midnight"?  "Half past midnight"?

I took a survey (n=1; I asked my wife) and she felt it would be more natural around midnight to say "quarter to twelve"/"quarter past twelve" etc.  So I think I'll keep "noon" and "midnight" only for those exact time points and use "twelve" otherwise.

Thoughts?

leskerr:
My wife and I both would say "quarter to 12" and "quarter/half past 12". So that's n=3 for your survey. :-)

AESilky:
My wife and I concur - so now you have n=5. :-)

Clock.py
```diff
@@ -30,61 +30,134 @@
...
+    return int(t / 3600)
+
+#
+# Convert seconds since midnight to the corresponding clock minute
+#
+def clock_minutes(t):
+    return int(int(t % 3600) / 60)
+
+#
+# Take an starting point and round up to the next interval-multiple:
```
AESilky, line 73: # Take a starting point and round up to the next interval-multiple:

Clock.py
```diff
@@ -30,61 +30,134 @@
...
+
+    clock_parser = argparse.ArgumentParser(parents=[config.PortOverride, config.SoundtOverride, config.WPMOverride])
+    clock_parser.add_argument("-b", "--begin", default=900, type=int, help="Beginning of time announcements ", metavar="time", dest="Begin")
+    clock_parser.add_argument("-e", "--end", default=2230, type=int, help="End of time announcements ", metavar="time", dest="End")
+    clock_parser.add_argument("-i", "--interval", default=60, type=int, help="The time announcement interval in minutes", metavar="minutes", dest="Interval")
+    args = clock_parser.parse_args()
+    
+    port = args.Port # serial port for KOB interface
+    speed = args.Speed  # code speed (words per minute)
+    sound = True if args.Sound == 'ON' else False
```
AESilky, line 133: Why to you need to do this? `config.Sound` is a boolean.

pwdirks: config.Sound is a boolean but args.Sound is a string ('ON' or 'OFF') and this converts 'sound' to be boolean, since even the string 'OFF' is interpreted as 'True' in an 'if' statement etc.  bool('ON') and bool('OFF') are both True'?

pykob/config.py
```diff
@@ -241,4 +245,14 @@ def readConfig():
 
 # ### Mainline
 readConfig()
+
+PortOverride = argparse.ArgumentParser(add_help=False)
+PortOverride.add_argument("-p", "--port", default=Port, help="The  name of the serial port to use ", metavar="portname", dest="Port")
+
+SoundtOverride = argparse.ArgumentParser(add_help=False)
+SoundtOverride.add_argument("-a", "--sound", default='ON' if Sound else 'OFF', choices=['ON', 'OFF'], help="'ON' or 'OFF' to indicate whether morse audio should be generated", metavar="sound", dest="Sound")
```
AESilky, line 253: So `Sound` can't be kept as a boolean? That seems like a problem. I don't like each application having to do 'ON'|'OFF' processing. Also, does `saveConfig` still work if an application wants to update the configuration?

Clock.py
pwdirks: Got it - fixed.

Clock.py
```diff
@@ -30,61 +30,134 @@
...
-        (1900, 'The time is seven oclock     L  L  L  L  L  L  L'),
-        (2000, 'The time is eight oclock     L  L  L  L  L  L  L  L'),
-        (2100, 'The time is nine oclock     L  L  L  L  L  L  L  L  L'),
-        (2200, 'The time is ten oclock     L  L  L  L  L  L  L  L  L  L')]
-
-    myKOB = kob.KOB(port=PORT, audio=SOUND)
-    mySender = morse.Sender(WPM)
+    log.log("Starting Clock")
+
+    clock_parser = argparse.ArgumentParser(parents=[config.PortOverride, config.SoundtOverride, config.WPMOverride])
```
AESilky, line 125: So, does every program that wants to use arguments other than the 'standard' arguments have to do this? 
I guess it's not too bad if we document it well.

pykob/config.py
pwdirks: Well, I'm fine with the argument parameter coming out as a string and this just has the user interpret args.Sound as a string.  I see config.py has setSound do a "_sound = strtobool(str(sound))" which converts 'OFF' to 'False' (instead of casting it with bool(str) which interprets 'OFF' as 'True'.  Are you suggesting I just use "strtobool" above instead?

Or do you want to be able to specify "-a ON" and "-a OFF" but have args.Sound come out as a boolean?  Is there an easy way to do that in the arg. parser?

Note that one alternative would be to allow (mutually exclusive) overrides '-a' to mean "audio ON" and, say, "-q" (quiet) for "audio OFF" and use action=store_true/store_false but I think I prefer the "-a ON" / "-a OFF" style.

Clock.py
pwdirks: I think of it as "Here's how you specify the standard, pre-configurable parameters that you want to be able to override".  You can omit overrides for things that don't apply to avoid getting into a situation where you end up accepting overrides for configuration parameters that don't even apply to the particular program?

pwdirks:
Les, I believe I've addressed everything you mentioned above - please have another look and I'll merge in this version.  You're welcome to tweak it further as you see fit then?

leskerr:
As usual, I can't keep up with you guys, and I don't want you to slow down on my account. Seriously...I mean it!

I have the latest version of Clock.py (I think!) running on my RPi. I have it set to announce the time on the quarter hour, and I love it! At some point I want to read through the code in detail and learn from the changes, especially the handling of parameters.

We talked a bit about the possible advantage for some users to specify Farnsworth timing for the Clock program. I think it makes more sense for the sender class in pykob/morse to default to the config settings for FW_CHAR_SPEED and FW_SPACING, if these aren't specified when a sender class is created. The user can set these parameters once using Configure.py, and the values will apply to all programs that use the code sender. I think there will rarely, if ever, be a need to do an application-specific override. I'll submit an Issue for this.

pykob/config.py
AESilky: I would use `strtobool` as that handles 'YES', 'TRUE', 'ON', '1' for `True` and 'NO', 'FALSE', 'OFF', '0' for `False`. So they could enter '--sound on', '--sound yes', '-a off', etc.

pwdirks:
Hi Les,

> On May 3, 2020, at 5:08 PM, Les Kerr <notifications@github.com> wrote:
> 
> 
> As usual, I can't keep up with you guys, and I don't want you to slow down on my account. Seriously...I mean it!
> 
> I have the latest version of Clock.py (I think!) running on my RPi. I have it set to announce the time on the quarter hour, and I love it! At some point I want to read through the code in detail and learn from the changes, especially the handling of parameters
> 
Wait until my latest check-in is complete and you’ll have the all-singing, all-dancing versions of Clock.py, complete with appropriate references to “midnight” and “a quarter to twelve” for those awake at the right time to notice :-)

Do have a look at the code, and let me know what you think generally and especially wrt. readability.  There are a lot more individual small functions now compared to the previous implementation which only had to run through a pre-defined list of hand-crafted announcements.  FWIW my usual rule of thumb is if I find myself writing the same operations twice in two different places I’ll define a function to do it right consistently - I never want to risk fixing up one place and leaving the same issue elsewhere.  In Clock.py that goes to consistently truncating hours, extracting hours/minutes from time values, formatting the time message, etc.  The other reason I sometimes add a function to do it is because it makes it a lot cleaner to apply some small amount of logic to an argument in a function call or computation where I can’t easily do the same in-place without some ugly temporary variable use or two-step calculation.

I’m new to Python, too - I welcome all comments!
> We talked a bit about the possible advantage for some users to specify Farnsworth timing for the Clock program. I think it makes more sense for the sender class in pykob/morse to default to the config settings for FW_CHAR_SPEED and FW_SPACING, if these aren't specified when a sender class is created. The user can set these parameters once using Configure.py, and the values will apply to all programs that use the code sender. I think there will rarely, if ever, be a need to do an application-specific override. I'll submit an Issue for this.
> 
Please do - I think we should end up with a setup where you can configure your favorite character speed/overall speed with Farnsworth spacing applied either between letters or between words and have the underlying code take care of generating everything that way consistently.

73, N6PWD
-Patrick.

Clock.py
AESilky: I would use `strtobool()`.

pwdirks: Done.  The latest change actually includes that.  So now it's just

    args = clock_parser.parse_args()
    ...
    sound = strtobool(args.Sound)

pwdirks:
Thanks for the review!

PR #36, opened 2020-05-11
Add KeyboardInterrupt guard in main code block of all programs

pwdirks:
I initially started trying to ensure a 'try: ... except:' block surrounded ALL executed statements, including 'import' but that got pretty messy so I abandoned that and reverted to changes ONLY in the main routine of a program.

I did in a few places separate out function definitions from their original location.

Comments welcome; if this looks OK I'll merge it later this week.  We can always tweak the code further if that's desirable.  This will NOT guarantee a ^C immediately on startup won't generate a traceback but should cleanly interrupt all main loops.

### DIFF
--- a/Clock.py
+++ b/Clock.py
@@ -191,4 +191,4 @@ try:
             time.sleep(24*3600 - now)  # wait until midnight and start over
 except KeyboardInterrupt:
     print()
-    sys.exit(0)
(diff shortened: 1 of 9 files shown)


leskerr:
You've been busy! I spot checked a few programs and didn't see any problems.

PR #38, opened 2020-05-20
Es 12 config opts

AESilky:
Add additional user preference configuration values.

This adds user configuration values for:
1. Character Speed (in WPM)
2. Sounder enable/disable (requires `port` to be configured, but allows turning the sounder on/off without needing to set `port` to 'NONE')
3. Spacing for Farnsworth timing (Character or Word)
4. Station to use by default
5. Wire to use by default

This just allows setting, recording (saving), and making these values available. It doesn't implement using them. That needs to be in separate issues (suggest one per value).

Configure (Configure.py) and Clock (Clock.py) were updated to process the new options. As noted, Clock does not do anything with the new options, but it supports including them on the command line when starting the application.


AESilky:
@pwdirks, @leskerr, please review and provide comments. Once we all approve, I will merge this in.

AESilky:
Added requested options from review.

AESilky:
Added requested options from review. Please review again.

leskerr:
1. There are a couple of typos in the help message: `pring` should be `print`, and `transmited` should be `transmitted`.
2. I would change `Set the minimum character speed in WPM (used with word speed for Farnsworth timing).` to `Set the minimum character speed in WPM (used for Farnsworth timing).` I know what `word speed` means from our discussions, but out of context it would not be understood.
3. I would change `Set the text speed in WPM (used with character speed for Farnsworth timing).` to `Set the overall code speed in WPM.` I don't see a need to mention Farnsworth here. And it probably makes more sense to show the `textspeed` parameter before the `charspeed` parameter.
4. I would change `Transmit the content to the configured wire-station.` to something like `Transmit over the internet on the specified wire.` See the following comment.
5. I would change `Set the Station to connect to.` to `Set the Station ID.` What's being set here is the user's own identification (e.g., office call, name, location).

Also, would it be possible for the `-i` option to include version numbers for Python, PyKOB, pySerial, and PyAudio? I know from experience this information is helpful when users need help with troubleshooting. I can submit this as a separate enhancement issue if that's the right thing to do.

AESilky:
From review by @leskerr - Fix typos, change help text, add 'code type' to `Configure`.

AESilky:
Make file comment match help and add processing of a missing option in Configure.

leskerr:
The help text looks good to me, but when I try the `-i` option I get a Python error.

```
C:\Users\Les\Documents\GitHub\PyKOB>configure -i
User: Les
User Home Path: C:\Users\Les
User Configuration File: C:\Users\Les\AppData\Local\pykob\config-les.ini
App Configuration File C:\ProgramData\pykob\config_app.ini
OS: nt
System: Windows
Version: 10
Platform win32
Host: Kerr-SurfacePro
  File "./Configure.py", line 220, in <module>
    main(sys.argv[1:])
  File "./Configure.py", line 198, in main
    if code_type:
UnboundLocalError: local variable 'code_type' referenced before assignment

C:\Users\Les\Documents\GitHub\PyKOB>
```

Clock.py
```diff
@@ -127,21 +126,29 @@ def announce(s, kob, sender):
...
+    arg_parser = argparse.ArgumentParser(description="Morse Cuckoo Clock", parents=\
+     [\
+      config.serial_port_override, \
+      config.code_type_override, \
+      config.sound_override, \
+      config.sounder_override, \
+      config.spacing_override, \
+      config.min_char_speed_override, \
+      config.text_speed_override])
+    arg_parser.add_argument("-B", "--begin", default=900, type=int, help="Beginning of time announcements ", metavar="time", dest="Begin")
```
pwdirks, line 140: Why switch to all-uppercase characters for options?  Especially when most of the command line is lower case commands (e.g. 'python3 Clock.py') is there anything wrong with using lowercase option characters, which in my experience are more common in UNIX/Linux commands?

Clock.py
```diff
@@ -127,21 +126,29 @@ def announce(s, kob, sender):
...
+    arg_parser.add_argument("-E", "--end", default=2200, type=int, help="End of time announcements ", metavar="time", dest="End")
+    arg_parser.add_argument("-I", "--interval", default=60, type=int, help="The time announcement interval in minutes", metavar="minutes", dest="Interval")
+    arg_parser.add_argument("-P", "--print", action='store_true', default=False, help="Whether to print text locally as it's sent to the sounder", dest="Text")
+    args = arg_parser.parse_args()
     
-    port = args.Port # serial port for KOB interface
-    speed = args.Speed  # code speed (words per minute)
-    if (speed < 1) or(speed > 50):
```
pwdirks, line 141: Why not check this (and bail out) before leaving 'speed' set to a bad value from args.Speed (although I do agree this is the right time to check, vs. Configure.py)

Clock.py
```diff
@@ -164,11 +171,11 @@ def announce(s, kob, sender):
...
         sys.exit(1)
     annc_interval = args.Interval * 60      # announcement interval (sec)
-    local_text = args.Text;
+    local_text = args.Text
     
     myKOB = kob.KOB(port=port, audio=sound)
-    mySender = morse.Sender(speed)
-    
+    mySender = morse.Sender(text_speed)
+
```
pwdirks, line 178: Is there a blank space at the start of line 178 (which is otherwise empty)?

Configure.py
```diff
@@ -67,51 +109,111 @@ def main(argv):
     # System configuration
     port = None
     # User preferences
+    local = None
```
pwdirks, line 112: Isn't it possible to set default values in the 'args' parameters, possibly even centrally, so no program should have to worry about initializing these variables but can instead always rely on 'args.xyz' (which IIRC will be 'None' anyway if no default value was supplied)?

AESilky:
The common preferences use '-t' for "text speed" so I needed to use
something else for 'type' so I used 'T' for it because I figured it would
be used less often. I tried to keep the more commonly used options lower
case and resort to upper case for the cases where the lower case option was
already in use.

I prefer the dash-dash options, since they use full words so they can be
descriptive and not conflict.

AESilky:
I will check this more closely when I'm on my computer, but in Configure.py
I want to know if an option was specified, as I only want to save if the
user specified one or more values.

For the display the values are read from the config object so that is what
is (has been) saved.

(from my phone)

pwdirks:
Hi Ed,

I think that makes sense for things like ‘Port’, for which no sensible default probably exists, and which should probably come back ’None’ if nothing’s configured and nothing specified on the command line.  OTOH for things like ’speed’ I think we can reasonably default to, say, 18 or 20 WPM and do something sensible without forcing the user to run configure.sh?

Regards,
-Patrick.

pwdirks:
OK, but why uppercase ‘begin time’ to ‘-B’ and ‘end time’ to '-E’?  Are there already -b and -e options defined for other purposes?

Thanks,
-Patrick.

AESilky:
> Hi Ed, I think that makes sense for things like ‘Port’, for which no sensible default probably exists, and which should probably come back ’None’ if nothing’s configured and nothing specified on the command line. OTOH for things like ’speed’ I think we can reasonably default to, say, 18 or 20 WPM and do something sensible without forcing the user to run configure.sh? Regards, -Patrick.

The Config object has default values for all of the parameters. It has a default of 18 for both min_char_speed and text_speed. So you will get those values from `config.char_speed_min` and `config.text_speed` if the user hasn't run Configure. You will get the values that are in the INI file if they exist.

The option-overrides in Config set a default from the value in the Config object, they should have a value based on the default or the value saved in the INI file (set by running Configure).

Maybe we should create an `action` class that the option-overrides use to set the specified values on the Config object directly, such that all of the 'standard' options (that are set as parents by the application) update the Config object and no action is required by the application for those options. The application would only processes the options it has added. Then the application pulls all of the config values from the Config object (rather than from the args).

Clock.py
AESilky: I didn't change how this was handled, I just changed the parameter names to match the new names. It does a `sys.exit(1)` if the value is out of range.

Maybe I'm misunderstanding your comment?

Clock.py
AESilky: I will switch Clock back to lowercase option characters.

Clock.py
AESilky: Yes. Thanks for catching that. I removed it.

Configure.py
AESilky: The default values are in the Config object. These are used if there hasn't been a value configured and saved using Configure.

Maybe I should create an 'action' object used in the argument overrides provided by the Config object. That way any common options specified would directly update the Config object. Then the application would read any needed Config values from the Config object (like config.text_speed, config.port, etc.) rather than reading them from the args. The only args the application would access/process are the options specific to the application.

AESilky: I will work on creating that and see how it works.

AESilky:
@leskerr I added you back as a reviewer (sorry), to review the addition of the 'system information' additions you requested (Python version, PyKOB version, PyAudio version, and PySerial version.

Please feel free to review/comment on anything/everything, but I'm requesting your review of the added information.

pwdirks:
> On May 24, 2020, at 12:33 AM, Ed Silky <notifications@github.com> wrote:
> 
> 
> Hi Ed, I think that makes sense for things like ‘Port’, for which no sensible default probably exists, and which should probably come back ’None’ if nothing’s configured and nothing specified on the command line. OTOH for things like ’speed’ I think we can reasonably default to, say, 18 or 20 WPM and do something sensible without forcing the user to run configure.sh? Regards, -Patrick.
> 
> The Config object has default values for all of the parameters. It has a default of 18 for both min_char_speed and text_speed. So you will get those values from config.char_speed_min and config.text_speed if the user hasn't run Configure. You will get the values that are in the INI file if they exist.
> 
> The option-overrides in Config set a default from the value in the Config object, they should have a value based on the default or the value saved in the INI file (set by running Configure).
> 
> 
That sounds sensible.  So you can always pick out ‘arg.xyz’ for the specified version, either as default (absent a config), as default for the argument (from the config) or as specified by the user (if overridden).
> Maybe we should create an action class that the option-overrides use to set the specified values on the Config object directly, such that all of the 'standard' options (that are set as parents by the application) update the Config object and no action is required by the application for those options. The application would only processes the options it has added. Then the application pulls all of the config values from the Config object (rather than from the args).
> 
> 
If I follow you correctly you’re proposing to side-effect the config object when I override an argument, just so that I can always reach for the ‘config.xyz’ value to get the desired value for a given run?  I think that would be very counter-intuitive - I don’t expect to change the default speed just because I, say, specify 25 WPM on a run.  The config should be what I configured, but I think that DOES make the perfect default for an argument override.

Hope that makes sense?

Thanks!
-Patrick.

AESilky:
It wouldn't save the values (unless the `save_config` method was used by the app), just set them on the config object. Probably makes more sense once we create a class for Config so there isn't a single shared instance. I'm envisioning that we will pass a Config object to `morse` and `kob` rather than passing a bunch of arguments. The Config object would be initialized from defaults and the INI values and overrides from the command line, plus any values the app wants to change, then it would be passed as one parameter.

That would free the apps from having to deal with any of the common configuration value processing in their options processing code - making it much easier to implement an application while still supporting local overrides of the common properties.

I will leave it as is for this check-in, but would like to discuss the approach.

Clock.py
AESilky: @pwdirks - Is this good as-is, or are you asking me to change something? Sorry, that I don't completely understand what you are asking/suggesting.

pwdirks:
Hi Ed - I think it’s more a stylistic thing and some of the code didn’t show the pattern well.  Rather than, say

	a = args.a
	if (a < 1) or (a > 50)
	    <print an error and exit>

I’d prefer

	if (args.a < 1) or (args.a > 50)
	    <print an error message and exit>
	a = args.a
	...

and of course the check on args.a could be done immediately following argument parsing, not right where it’s used.  But nothing functionally wrong with the code as-is and no need to change the code or defer the merge.

Regards,
-Patrick.

Configure.py
AESilky: @pwdirks - Per the conversation in the issue comments, I will leave this as is for now and we can discuss approaches for later.

Currently, the values are set by:  
1. Defaults coded in the Config object
2. Values configured by the user using Configure.py (or manually editing the INI file)

These values are then used to initialize the values for the command line option processing.

Does that explaination resolve this, or is there something else that you are suggesting?

AESilky:
Okay. I thought you wrote that - but I'm happy to change it as you suggest.

Clock.py
pwdirks: [ looks like my original response by e-mail didn't properly get attached here.  Repeating: ]

Hi Ed - I think it’s more a stylistic thing and some of the code didn’t show the pattern well.  Rather than, say

	a = args.a
	if (a < 1) or (a > 50)
	    <print an error and exit>

I’d prefer

	if (args.a < 1) or (args.a > 50)
	    <print an error message and exit>
	a = args.a
	...

and of course the check on args.a could be done immediately following argument parsing, not right where it’s used.  But nothing functionally wrong with the code as-is and no need to change the code or defer the merge.

Configure.py
pwdirks: [ I think another emailed response was lost.  Repeating here: ]

The Config object has default values for all of the parameters. It has a default of 18 for both min_char_speed and text_speed. So you will get those values from config.char_speed_min and config.text_speed if the user hasn't run Configure. You will get the values that are in the INI file if they exist.

The option-overrides in Config set a default from the value in the Config object, they should have a value based on the default or the value saved in the INI file (set by running Configure).


That sounds sensible.  So you can always pick out ‘arg.xyz’ for the specified version, either as default (absent a config), as default for the argument (from the config) or as specified by the user (if overridden).
Maybe we should create an action class that the option-overrides use to set the specified values on the Config object directly, such that all of the 'standard' options (that are set as parents by the application) update the Config object and no action is required by the application for those options. The application would only processes the options it has added. Then the application pulls all of the config values from the Config object (rather than from the args).


If I follow you correctly you’re proposing to side-effect the config object when I override an argument, just so that I can always reach for the ‘config.xyz’ value to get the desired value for a given run?  I think that would be very counter-intuitive - I don’t expect to change the default speed just because I, say, specify 25 WPM on a run.  The config should be what I configured, but I think that DOES make the perfect default for an argument override.

Hope that makes sense?

PR #52, opened 2020-05-26
Change MKOB title bar and file extension

leskerr:
Changed MKOB 1.0.0 to MorseKOB 4.0.0 in title bar and setup.py.  Also added license to setup.py.

It seems like a bad idea to have the version number in two separate places (MKOB.py and setup.py) but I don't know how to avoid this. If anyone has a better way we can fix it now, or I can open an issue to be dealt with later.

### DIFF
--- a/MKOB/MKOB.py
+++ b/MKOB/MKOB.py
@@ -25,11 +25,9 @@ SOFTWARE.
 """
 
 """
-
-MKOB.pyw
+MKOB.pyw  (Why does this say MKOB.pyw instead of MKOB.py? I don't remember. @leskerr)
 
 Python version of MorseKOB 2.5
-
 """
 
 import tkinter as tk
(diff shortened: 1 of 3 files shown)


MKOB/MKOB.py
```diff
@@ -25,19 +25,17 @@
 """
 
 """
-
-MKOB.pyw
+MKOB.pyw  (Why does this say MKOB.pyw instead of MKOB.py? I don't remember. @leskerr)
```
pwdirks, line 28: Running a ".py" program with "python.exe" executes in a terminal window, which stays open while the program runs.  Changing too ".pyw" causes it to be run by pythonw.exe by default instead.  This suppresses the terminal window on startup, which may be more appropriate for a program that presents a GUI.

I think for the main "MKOB" program you probably want ".pyw".

leskerr: Thanks, @pwdirks. I changed the filename to MKOB.pyw. It seems to work either way with Windows. I'm not able to test it with Linux right now.

AESilky:
Can this be closed?

leskerr:
> Can this be closed?

Hopefully it's closed now. Let me know if there's still a problem.

PR #61, opened 2020-05-30
Lk 59 remove python 2 code

leskerr:
Tidy up loose ends related to obsolete Python versions


pykob/internet.py
```diff
@@ -91,9 +91,8 @@ def read(self):
                 pass  # ignore Ack packet
             elif nBytes == 496:  # code or ID packet
                 self.tLastListener = time.time()
-                # stnID, seqNo, *code = codePacketFormat.unpack(buf)  # works in Python 3.x
```
pwdirks, line 94: Did you mean to use this style instead here, then, now that it's strictly Python 3-only?

leskerr: Good eye, @pwdirks, nothing gets past you!

Actually, I did try the Python 3-only approach, which I used to think was cleaner, but it failed with a list/tuple conflict. I could have dealt with that in turn, but it got a bit messy.

As long as it's working I'll leave it for someone with a better eye for style to take care of.

Thanks for the review!

pwdirks: Fair enough.  Might want to add a question mark, then, or "... is supposed to work as well in Python 3?", or just leave as-is.  Your call.

Clock.py
```diff
@@ -24,7 +24,8 @@
 SOFTWARE.
 """
 
-"""Clock.py
```
AESilky, line 27: Following the """ immediately with the file/class/method name ("""Clock.py) is done to provide Python documentation from the docstring comments. See [PEP 257](https://www.python.org/dev/peps/pep-0257/), [Documenting Python Code](https://realpython.com/documenting-python-code/#documenting-your-python-code-base-using-docstrings) and [Documenting Python](https://devguide.python.org/documenting/)

Time.py
```diff
@@ -23,7 +23,9 @@
 OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE.
 """
-"""Time.py
```
AESilky, line 26: See previous review comment about Python documentation docstrings (in `Clock.py`).

Clock.py
leskerr: Thank you! I didn't know that. I thought they were typos. I'll fix them before merging.

leskerr:
Thanks for the review, guys. As you know, I have a lot to learn. I appreciate every bit of feedback I can get!

Clock.py
AESilky: Perfect. All good then.

PR #67, opened 2020-06-10
Lk 57 fix reader window jitter

leskerr:
I've already spent way more time than I should have fiddling with the MorseKOB GUI widget attributes, without a whole lot to show for it. I found the 'themed Tk' widgets to be poorly documented, complicated to use, and the end result didn't look like an improvement. Therefore, for now at least, I'm sticking with tried and true Tkinter widgets.

Switching to the `tkinter.scrolledtext` module did get rid of the serious jitter problem in the code reader window, so that made it all worthwhile. I still see an occasional flicker in that window, but nothing like it was before. I'm ready to move on.

My philosophy here, by the way, is to keep the look-and-feel of the main MKOB window as close as possible to MorseKOB 2.5 (on Windows, at least). I want to make it comfortable for users to transition to MorseKOB 4.0, and any changes—even if they're improvements—can be jarring at first.

I have no misgivings about following up later with a new improved version, maybe with a different screen layout, possibly even a new name for the program if that seems best.

Here's what this version looks like to me on Windows and Linux. I'd appreciate it if you could send me screenshots of what it looks like on the Mac. And of course, any comments are always welcome.

MKOB 4 0 1 screenshot (Windows)

MKOB 4 0 1 screenshot (Linux)


AESilky:
MKOB on my Mac:

image

Note that my console window stays open and does so even after I close the MKOB window.

MKOB/MKOB.pyw
```diff
@@ -24,19 +24,25 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE.
 """
 
-"""MKOB.pyw
```
AESilky, line 27: I know it doesn't look great - but following the `"""` immediately by the module name is used to generate documentation (at the point we decide to do that). If there is a line break, the doc processing won't pick up the module name for the documentation.

MKOB/MKOB.pyw
```diff
@@ -24,19 +24,25 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
...
+"""
+MKOB.pyw
 
 Python version of MorseKOB 2.5
+
+MorseKOB 4.0.1  2020-06-10
+- fix jitter in code reader window
```
AESilky, line 33: I'm fine with this for now, but at some point the 'Version Info' should be auto generated into a 'versions.md' (or similar) file, not manually included into the source file.

MKOB/kobwindow.py
```diff
@@ -182,6 +175,22 @@ def __init__(self, root, VERSION):
...
+    for widg in all_widgets:
+        print("\nWidget Name: {}".format(widg.winfo_class()))
+        keys = widg.keys()
+        for key in keys:
+            print("Attribute: {:<20}".format(key), end=' ')
+            value = widg[key]
+            vtype = type(value)
+            print("Type: {:<30} Value: {}".format(str(vtype), value))
+
 ## TAKE CARE OF THESE
```
AESilky, line 194: If this is important - can you use `TODO:` as a lead-in so we can consistently search for it (please many IDE's automatically highlight/index those contents).

leskerr:
> MKOB on my Mac:
> 
> Note that my console window stays open and does so even after I close the MKOB window.

Can you open this as a new issue for the Mac? Maybe you or @pwdirks can look into it sometime.

MKOB/MKOB.pyw
leskerr: > I know it doesn't look great - but following the `"""` immediately by the module name is used to generate documentation (at the point we decide to do that). If there is a line break, the doc processing won't pick up the module name for the documentation.

Sorry about that. I can fix it.

When I read the `docstring` documentation that you pointed me to, it claimed that it works equally well if the module name is on the next line, but I never did try it to see what happens.

MKOB/MKOB.pyw
leskerr: Ah, you beat me to it! I was planning to raise the whole question of version control. I need to be educated on best practices in this area.

MKOB/kobwindow.py
leskerr: Thanks for the suggestion! I'll follow that convention in the future.

MKOB/MKOB.pyw
AESilky: Ah, okay... Maybe it has changed since I used it. Leave it as is (if you haven't changed it) and we can try it.

PR #70, opened 2020-06-13
Processing of saved config values of 'STATION' and 'WIRE'.

AESilky:
@leskerr identified a problem in reading the 'station' and 'wire' parameters. Inspection showed that the values were saved correctly in the configuration file, but were not being read back correctly.

This fixes the processing of the '_STATION' and '_WIRE' key so that values of "NONE" or specific values are handled correctly.


leskerr:
It's working fine now. Thanks!

Issue #55, opened 2020-05-27
MKOB: Sender doesn't send code


AESilky:
When connected to a wire, using the (un-labeled) lower left text box to enter text to send doesn't send code.

Steps to reproduce:
1. Run MKOB
2. Connect to a wire (10 was used for the test)
3. On a different machine run MorseKOB 2.5 (Windows app)
4. Connect to the same wire using a different station name
5. In MKOB open the circuit
6. Turn 'Code Sender' 'On'
7. Enter text in the (un-labelled) lower left text box
8. Monitor MorseKOB 2.5 (from step 3.)
9. Note that no code is received

PR #80, opened 2020-06-24
Lk 54 station list

leskerr:
The station list seems to be working now, although it hasn't been thoroughly tested. Please report any unexpected behavior, even if it seems minor.

This issue will have to be revisited when we implement the code sender window (#55).


AESilky:
Note - in my testing the station name indicates the platform:
'M' = Mac
'W' = Windows
'L' = Linux

PR #86, opened 2020-06-29
Lk 55 implement code sender

leskerr:
The keyboard sender window is working, but it's not fully integrated with the other MKOB components yet. For now, you have to enter a tilde (~) at the beginning of a transmission and end it with a plus sign (+).

### DIFF
--- a/MKOB/kobactions.py
+++ b/MKOB/kobactions.py
@@ -25,26 +25,20 @@ SOFTWARE.
 """
 kobactions.py
 
-Defines actions for MKOB
+Handle actions for controls on main MKOB window
 """
 
 import tkinter.messagebox as mb
+import time
 from pykob import config, kob, internet, morse
 import kobconfig as kc
+import kobmain as km
 import kobstationlist as ks
-import time
 
 import pykob  # for version number
 print("PyKOB " + pykob.VERSION)
 
-# global variables
-
 kw = None  # initialized by KOBWindow
-mySender = None
-myReader = None
-myInternet = None
-##running = True
-connected = False
 
 # File menu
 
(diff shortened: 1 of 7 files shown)


MKOB/kobactions.py
```diff
@@ -25,34 +25,28 @@
...
 import tkinter.messagebox as mb
+import time
 from pykob import config, kob, internet, morse
 import kobconfig as kc
+import kobmain as km
 import kobstationlist as ks
-import time
 
 import pykob  # for version number
 print("PyKOB " + pykob.VERSION)
```
AESilky, line 39: Even though it's in the MKOB title bar - since you are printing this in the console it would be nice to print the MKOB version in the console as well.

MKOB/kobmain.py
```diff
@@ -0,0 +1,95 @@
...
+import kobkeyboard
+
+mySender = None
+myReader = None
+myInternet = None
+connected = False
+
+kob_latched = True
+keyboard_latched = True
+internet_latched = True
```
AESilky, line 44: Is there a state table (or something) that explains what the combination of these three states mean? I see from the code that if two are true the third is set false and an action is performed.

MKOB/kobactions.py
leskerr: > Even though it's in the MKOB title bar - since you are printing this in the console it would be nice to print the MKOB version in the console as well.

No problem. I've made the change and it'll show up in the next pull request.

MKOB/kobmain.py
leskerr: > Is there a state table (or something) that explains what the combination of these three states mean? I see from the code that if two are true the third is set false and an action is performed.

Those latch variables indicate whether the respective system components are latched in their circuit closed state. For the key, that means the key's circuit closer is closed. For the keyboard, it means the user has entered a plus sign to close the circuit. For the internet, it means the special mark code of +1 has been detected, meaning the remote station's key is closed.

This all touches on the pesky 'break behavior' issue, which I'm struggling with right now. The edge conditions (which should rarely happen, but always seem to anyway) are driving me nuts! There are so many possible interactions. Once I get this working, I will document it. If for no other reason, so I can understand it later.

AESilky: That really helps - thanks! I'll think about it with you.

PR #88, opened 2020-07-02
Es 74 receive2

### LINKED ISSUE #74: MKOB: Received code stops short of that sent
The received code window stops two characters short of what is sent on the wire.

For example, when receiving code on a wire where another station sends "ABCDEFG", the received code window shows "ABCDE" (missing the "FG").

Image from MKOB-2.5 (sending) and MKOB-4.0.3 (receiving):
6-19-2020 10-46-08 PM

AESilky:
Use a timer to flush the receiver if no code is received in 8 'true dot' times. This flushes the last 1-2 buffered characters out if nothing further is received.

Fixes #74 

Tests:
- Send from MKOB-2.5 and receive using 4.0 on a different machine. Notice that the complete message is received/displayed in a timely manner.


leskerr:
@AESilky Thanks for adding the timer to `pykob.morse`. It works great on my tests so far on machine sent code. I have a feeling 8 dot lengths will turn out to be too short with Morse sent by real operators. Very often, when someone's sending with a bug the dots will be significantly faster than the overall speed of the Morse. Where this comes into play is that the internal space of the 'Morse' letters can be more the 8 dot lengths. In this case, the internally spaced letters would be incorrectly decoded as two separate letters.

I'd rather merge this branch now and deal with any tweaking of the timer later. I went ahead and did the merge myself. I hope I wasn't out of line in doing this on my own initiative. Feedback welcome!

AESilky:
@leskerr great insight! I just thought that basing it off of dot speed was better than a fixed value (say 0.8 seconds). But I took a guess as what percentage of 'dot-speed' would be reasonable.

leskerr:
> @leskerr great insight! I just thought that basing it off of dot speed was better than a fixed value (say 0.8 seconds). But I took a guess as what percentage of 'dot-speed' would be reasonable.

@AESilky Your approach is exactly right. The delay should indeed be a multiple of the currently received dot length. When we get on the wire together I can send on my bug and we'll see what actually happens. Maybe even make some adjustments on the fly.

AESilky:
I look forward to the live code (even though I can't contribute (I can key "THE" semi-reliably)), but I would love to hear some 'real' code.

AESilky:
"THE" appears twice in 'The quick brown fox...' and it is fairly simple and common. :-)

leskerr:
Yup, 'the' and 'and' occur so frequently, pretty soon you start hearing them as entire words rather than individual letters.

leskerr:
I stand corrected. Eight dot lengths for the timer is just fine. I tried sending with my bug, and even when I exaggerate the spacing in my sending it decodes beautifully. No need to change anything. Nice job!

AESilky:
Thanks for confirming. I did think quite a bit about it (leaning on some of my Watson STT (speech to text) work), but (being a Morse novice) I did have to 'take a best guess' at it.

PR #89, opened 2020-07-02
Ignore generated  files and IDE files.

AESilky:


leskerr:
@AESilky Thanks for doing this. I knew it needed to be done, but I wasn't sure how to go about it.

PR #90, opened 2020-07-03
Lk 58 implement break behavior

leskerr:
All the major pieces of the program seem to be working, although I have very little confidence that the integration is bug-free. The `kobmain` module is a dark and scary place! I'm at the point where I'd like to start putting MKOB through its paces and capturing any bugs we find.

Remember, if you're using the keyboard, you need to start your transmissions with a tilde and end with a plus sign.

I'd like to demonstrate the break behavior sometime when the three of us can get on the wire together. Let me know when would work for you and I'll make myself available. Thanks!


MKOB/MKOB.pyw
```diff
@@ -35,6 +35,7 @@ import sys
 import kobwindow
 
 VERSION = "MorseKOB 4.0.6"
```
AESilky, line 37: I probably should have commented on this earlier (since this wasn't a current change), but I think 'VERSION' should just be '4.0.6'. The 'MorseKOB' should be added when printing/displaying it.

leskerr: Good suggestion. I'll change it the next time I touch the code.

PR #95, opened 2020-07-04
Lk 92 config settings

leskerr:
This pull request includes the following changes:

- use config settings for code type, Farnsworth spacing, local, and remote
- fix VERSION to include just the version number
- gray out unimplemented checkboxes
- decode long dash as T (not L) in International
- create a skeleton list of items to be described in MKOB 4.1 release notes


MKOB/MKOB.pyw
```diff
@@ -34,8 +34,8 @@ import tkinter as tk
 import sys
 import kobwindow
 
-VERSION = "MorseKOB 4.0.6"
-print(VERSION)
+VERSION = "4.0.7"
+print("MorseKOB " + VERSION)
```
AESilky, line 38: Just a note:
This is fine... I think you can also use the form `print("MorseKOB", VERSION)`

MKOB/kobconfig.py
```diff
@@ -38,20 +38,23 @@
 
 WPM            = config.text_speed      # overall code speed
 CWPM           = config.min_char_speed  # minimum character speed (Farnsworth)
-Spacing        = 1 if config.spacing == 'WORD' else 0
+Spacing        = 1 if config.spacing == config.Spacing.word else 0
                                         # 0: character, 1: word (Farnsworth spacing)
-CodeType       = 1 if config.code_type == 'INTERNATIONAL' else 0
+CodeType       = 1 if config.code_type == config.CodeType.international else 0
                                         # 0: American, 1: International
```
AESilky, line 44: Seeing this in use, I think we should change the definition of these enums so the 'name' is the text and the 'value' is the number. That way you can just use `config.spacing.value` and `config.code_type.value` rather than having to do this test-&-conversion.

As long as we establish a standard for these, it will make it easier for all applications that use the values.

I will create an issue, make that change, and we can use that here.

MKOB/kobwindow.py
```diff
@@ -44,7 +44,7 @@ def __init__(self, root, VERSION):
         self.root = root
         root.rowconfigure(0, weight=1)
         root.columnconfigure(0, weight=1)
-        root.title(VERSION)
+        root.title("MorseKOB " + VERSION)
```
AESilky, line 47: I stand by my suggestion that `VERSION` just be the version 'number', but seeing how many times this combination is used, it might be good to create a constant for it somewhere in the main mkob (`MKOB_VERSION_TEXT = "MorseKOB " + VERSION`) so you can use it consistently throughout.

MKOB/kobwindow.py
```diff
@@ -124,6 +124,7 @@ def __init__(self, root, VERSION):
         self.varCircuitCloser = tk.IntVar()
         chkCktClsr = tk.Checkbutton(lfm1, text='Circuit Closer',
                 variable=self.varCircuitCloser)
+        chkCktClsr.config(state='disabled')  # temporary
```
AESilky, line 127: I prefer a comment of:
 "# TODO: temporary". 
just because most development tools will highlight/list 'TODO' commented lines. The other option is that we add 'temporary' as a comment keyword for the tools to list.

An example is that I can bring up a list of all 'TODO' lines in my IDE (VS Code). We can also automatically generate a list of 'TODO' lines as part of a check-in if we enable some of the optional validation in GitHub (was waiting to do that until we are further along).

AESilky: As a note...
On the teams I worked on we used `TODO:` for something that should be addressed/done 'at some point' but didn't necessarily need to be done for the current release. We used `ZZZ:` for something we wanted to check in to be tested (for primary functionality) but needs to be addressed before being accepted for a release.

So, when doing a final check of code for a release, code with `TODO:` can be accepted, but code with `ZZZ:` needs to be addressed.

MKOB/kobconfig.py
AESilky: I created #97 to adjust the enums. I think this is good 'as-is' (for now) and we can adjust to make this simpler for this (and other applications) when we implement the change for #97.

MKOB/kobwindow.py
AESilky: Let me know if you think this is worth a change. If you don't think so, I'm fine - maybe something to consider for the future (or you disagree - fine as well).

MKOB/kobwindow.py
leskerr: I made this change. Please keep reminding me if I forget in the future.

MKOB/kobwindow.py
leskerr: Good suggestion. We might want to change the program name in the future. Best to have it in one place.

I made the change.

PR #100, opened 2020-07-08
Adjust config enums to have numeric values. Use enum names to store t…

AESilky:
…he configuration.

PyKOB/MKOB uses numeric values for the code type and the Farnsworth spacing. Rather than testing the enum and setting a numeric value, the enums are adjusted to have the value be the numeric value. Where a character value is needed (load/save configuration and command line options) the enum.name value is used.

### DIFF
--- a/MKOB/kobconfig.py
+++ b/MKOB/kobconfig.py
@@ -38,10 +38,8 @@ Audio          = config.sound           # simulated sounder on or off
 
 WPM            = config.text_speed      # overall code speed
 CWPM           = config.min_char_speed  # minimum character speed (Farnsworth)
-Spacing        = 1 if config.spacing == config.Spacing.word else 0
-                                        # 0: character, 1: word (Farnsworth spacing)
-CodeType       = 1 if config.code_type == config.CodeType.international else 0
-                                        # 0: American, 1: International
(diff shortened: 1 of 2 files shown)


pwdirks:
One thought about choice of enum values: I've had a longstanding practice of making "0" be an "invalid" value for an enum so that variables accidentally used without being set will not look like legitimate values if they're allocated zeroed.

Is it worth a separate follow-up issue to make sure all enum values are non-zero and/or define "0" as an "invalid" value?

leskerr:
> One thought about choice of enum values: I've had a longstanding practice of making "0" be an "invalid" value for an enum so that variables accidentally used without being set will not look like legitimate values if they're allocated zeroed.
> 
> Is it worth a separate follow-up issue to make sure all enum values are non-zero and/or define "0" as an "invalid" value?

That sounds like a good practice. I wish I'd thought of that when I assigned the integer values way back when. I don't think it's too late to change it.

AESilky:
I'm fine with that. I was just trying to make as little change as possible
to Les' existing code.

AESilky:
Simple to change in `config`. As I said, I picked values that matched the
existing code.

PR #101, opened 2020-07-09
lk-79-proportional-spacing

leskerr:
The idea behind proportional spacing in the code reader window is to have the space between characters reflect the actual timing of the sent or received code. Normally spaced characters have a narrow space inserted between them (about half the width of a normal space). Very closely spaced characters have no space between them. Pauses between characters are indicated by an amount of space proportional to the length of the pause. This is all so the user has the information necessary to interpret what's going on when hand sent code is not decoded properly by the code reader algorithm.

I know the spacing of the characters may look odd at first, but this is how MorseKOB 2.5 works and some operators (myself included) have come to appreciate it. We could make this an option in the future, especially as AI is incorporated into the decoding algorithm.

Here's how it looks with Windows on machine sent code. This screenshot also shows an instance of a dropped packet (the B in PUBLIC). This only happens rarely, so I was lucky to get an unplanned test of that feature.

Dropped packet

I also tried sending some Morse on my bug. First I sent a passage from Science magazine, making an effort to send accurate code. Then I sent some stream of conciousness stuff in a more cavalier style. There are multiple sending errors, that are my fault. Some of them I corrected, after sending a question mark which is the convention for when you're repeating something you've already sent. Often, though, operators simply pause and resend without including the (more formal) question mark. I did that also, and the pause is shown by the longer space between words.

Here's that screen shot. I see a lot of my T's are longer than the decoding algorithm expects and they get displayed as L's. The algorithm needs to be tweaked to make it more forgiving with this style of sending (which I think is fairly common among bug users).

Manual sending

I don't have a Mac to use for testing, so I'd really appreciate knowing how the screen looks on a Mac. I'm especially interested in whether the word wrap mechanism behaves properly. Lines should only be broken at word boundaries and not at the small space that's inserted between letters within a word (I'm using Unicode 'narrow no-break space' characters for this purpose). You can check that behavior by connecting to a wire with a machine sent feed, and then resizing the reader window to make it wider and narrower and see what happens with the word breaks.

I have tried this version of MKOB on my Linux laptop. The screen looks okay, but I'm having a problem running MKOB on Linux. The first time I run it in a terminal window, I get this error:

```
PyKOB 1.2.6
MorseKOB 4.0.8
Exception in thread Thread-3:
  File "/usr/lib/python3.8/threading.py", line 932, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.8/threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "/home/les/PyKOB/MKOB/kobkeyboard.py", line 47, in keyboard_send
    kw.txtKeyboard.tag_config('highlight', background='gray75',
AttributeError: 'KOBWindow' object has no attribute 'txtKeyboard'
```

If I run it a second time in the same terminal window, it works fine. If I try again later, however, the problem reappears. I'm not asking for help (yet) in fixing this problem, but I wanted you to know that I'm aware of it.


MKOB/kobmain.py
```diff
@@ -136,12 +138,26 @@ def update_sender(id):
 
 def readerCallback(char, spacing):
     """display characters returned from the decoder"""
-    n = int(spacing + 0.5)
-    if n > 5:
-        sp = "__" if char != "~" else ""
+    if kc.CodeType == 0:  # TODO: use config enum for American
```
pwdirks, line 141: I believe that change was merged; please use "CodeType.american" here now?

MKOB/kobmain.py
```diff
@@ -136,12 +138,26 @@ def update_sender(id):
...
+    elif sp < 0.8:
+        txt = NNBSP + " "
     else:
-        sp = n * " "
-    ka.codereader_append(sp + char)
+        n = int(sp - 0.8) + 2
+        txt = n * " "
+    txt += char
+    ka.codereader_append(txt)
+##    print("kobmain: {:.2f} '{}'".format(spacing, txt))  # TODO: temporary
```
pwdirks, line 160: Leftover debug code?  Might as well get rid of it now?

leskerr:
> Two minor comments - no objection to merging as-is.
> 
> Any way to omit the extraneous binary files etc. from your commit?

@pwdirks Thanks for the comments!

I don't see any binary files in this branch on the server. Which ones are you referring to?

AESilky:
I added the `dist` folder to `.gitignore` and removed the `dist` folder from the repo. I believe the 'D' on `dist` is probably the result of removing it. We'll see.

MKOB/kobmain.py
leskerr: I totally agree. I was aware that the new config had been merged on the server, but my branch was based on an earlier version of Master and the new config changes weren't reflected in my working directory. I'm sure there's a way I could have merged the new config into my branch, but I wasn't sure how to do it.

I will definitely change to CodeType.american the next time I touch the code (if someone else doesn't get to it first). I know that needs to be done before the enum values get changed, per your earlier suggestion.

MKOB/kobmain.py
leskerr: You have a sharp eye! Yes, that was debugging code. I almost deleted it before doing the commit, but I suspect I'll be revisiting the proportional spacing issue again and will likely want to reenable that statement. I'll be sure to clean it up when I'm done with it.

My plan is to use this version of MKOB on wire 11 during this Saturday's net and see how everything works. Don't be surprised if you see a slew of new issues get posted as a result!

All in all, I'm very happy with the way things are coming along. Having said that, I know there's a bunch of fine tuning that we'll want to do.

MKOB/kobmain.py
AESilky: I'll pull this branch, merge, make this change, and push it back.

MKOB/kobwindow.py
```diff
@@ -88,14 +88,14 @@ def __init__(self, root, MKOB_VERSION_TEXT):
 
         # reader
         self.txtReader = tkst.ScrolledText(frm1, width=40, height=15, bd=2,
-                wrap='word', font=('Helvetica', -14))
+                wrap='word', font=('Arial', -14))
```
AESilky, line 91: Any reason you switched to 'Arial'? Though it is pretty well accepted now, it is (actually) a Windows font. 'Helvetica' is a generic font.

leskerr: Believe it or not, Helvetica is not installed on my Windows 10 laptop.

Helvetica

I'm open to suggestions for a suitable font that comes preinstalled on all three platforms. The font needs to support the Narrow No-Break Space character at least, and ideally Thin Space and Hair Space as well.

AESilky:
I'm working on updates to use the enums (per @pwdirks  request). This isn't working correctly, but I wanted to push so Les can help.

AESilky:
This has been merged with the current master, so you can use the enum values.

MKOB/kobmain.py
AESilky: When/what is "Saturday's net"?

I would like to join if possible.

AESilky: Maybe (hopefully) I'll have an installer by then. It's kind of tough work (creating and then testing on Win/Mac/LX).

AESilky:
I made an attempt to convert everything in MKOB-PyKOB to using the enums.
@leskerr my code window shows 'gibberish' - so I'm sure I missed a `code_speed` use somewhere. Can you take a look (since you are much more familiar)?

AESilky:
@leskerr - if you `pull` you should get all of my changes and be able to test.
I'm happy to fix things if you point them out.

AESilky:
Worked with @leskerr realtime and resolved the enum issues to get things working. Enums now have non-zero values and everything appears to be working.

MKOB/kobwindow.py
AESilky: Since I have all three platforms, I'll look at fonts.

In our working this afternoon, I took your changes, so we can go with that for now. I think Mac maps 'Arial' to a similar (installed) font. Linix might do the same (haven't tested yet), but I don't like using 'Arial' in the code.

I'm good for now. I don't want to hold this up on a font issue.

PR #103, opened 2020-07-12
Encoding file index error.

AESilky:
Fix an index error that wasn't found the other day in the pair coding session.


leskerr:
@AESilky That fixed the problem. Thank you!

PR #106, opened 2020-07-12
Es 105 print text

AESilky:
Added option to print characters as they are encoded. Used the option in `sample.py`.

### DIFF
--- a/pykob/morse.py
+++ b/pykob/morse.py
@@ -80,8 +80,11 @@ class Sender:
             self.wordSpace = 0
         self.space = self.wordSpace  # delay before next code element (ms)
         
-    def encode(self, char):
+    def encode(self, char, printChar=False):
         c = char.upper()
+        if (printChar):
+            print(c, end="")
+            sys.stdout.flush()
         code = ()
         cti = 0 if self.codeType == config.CodeType.american else 1
         if not c in encodeTable[cti]:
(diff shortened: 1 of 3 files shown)


leskerr:
In `pykob/morse.py`:

```
        if (printChar):
            print(c, end="")
            sys.stdout.flush()
```

Python 3 provides an option on the print statement to flush the output.

```
        if (printChar):
            print(c, end="", flush=True)
```

Is there a reason not to use that form here?

And, a 'metacomment', can someone show me how to comment on a particular code snippet in a pull request? It looks so cool when you guys (@pwdirks and @AESilky) do it.

AESilky:
Thanks for reminding me of that. I had forgotten. I can change it when I'm
back in the code.

pykob/morse.py
```diff
@@ -80,8 +80,11 @@ def __init__(self, wpm, cwpm=0, codeType=config.CodeType.american, spacing=confi
             self.wordSpace = 0
         self.space = self.wordSpace  # delay before next code element (ms)
         
-    def encode(self, char):
+    def encode(self, char, printChar=False):
         c = char.upper()
+        if (printChar):
+            print(c, end="")
+            sys.stdout.flush()
```
leskerr, line 87: This is just a test comment. Please ignore.

PR #116, opened 2020-07-16
Local and Remote config default to 'ON'

AESilky:
@leskerr suggested that having the default be `On` for 'Local' and 'Remote' for a new installation would be helpful for new users.

### DIFF
--- a/pykob/config.py
+++ b/pykob/config.py
@@ -112,8 +112,8 @@ serial_port = None
 
 # User Settings
 code_type = CodeType.american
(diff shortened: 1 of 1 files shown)


AESilky:
Tested by removing my `.ini` file and then running MKOB, then checking the configuration manually and with `configure`.

This is a simple change so I'm merging without a review.

leskerr:
Thank you! Haven't tried it yet, but looks good. Please merge.

PR #122, opened 2020-07-25
Es 73 mkob installer

AESilky:
There is still work to do, but I wanted to get this in as a first step to get the code changes I had to make into master.

I need to create scripts for Mac and Linux to generate bundles for those platforms. There will also be changes down the road to run without the console and a few other things.


pwdirks:
Looks good to me!  I fired it up under Windows-32, connected to wire 107 and wire 101, heard the morse, saw the decoding; MKOB4 looks like it works fine.  I can't figure out what the "Configure" application was supposed to do - when I double-clicked on it it briefly flashed a terminal screen and quickly disappeared.  Is it supposed to be a GUI app?  Or invoked some other way?

leskerr:
Looks good to me. See my question about whether the `Configure` directory is necessary, since `Configure.exe` is included in the `MKOB` directory.

AESilky:
@pwdirks the Configure.exe runs Configure.py to allow you to view and set configuration values. It needs to be run from a command prompt. I'll see if I can figure out how to keep the command window open if you just double-click it.

@leskerr I don't see the comment you reference, but I didn't realize I was including the Configure folder in the zip. I'll adjust that so it's not included. It gets created in order to bundle the Configure app, then I copy Configure.exe into the MKOB folder (since all of the dependencies are already there, but I'm accidentally grabbing both when I zip.

AESilky:
I changed the zip step to not include the Configure folder. I also changed it so the MKOB contents go directly into the zip (no MKOB folder in the zip).

I put updated zip files in the OneDrive folder.

AESilky:
I'm not going to delete the branch because I'm using it to work on the Mac bundle.

leskerr:
> I changed the zip step to not include the Configure folder. I also changed it so the MKOB contents go directly into the zip (no MKOB folder in the zip).
> 
> I put updated zip files in the OneDrive folder.

@AESilky This works fine. Thank you!

PR #124, opened 2020-08-02
Updated to new pySerial 3.0 interfaces

pwdirks:


pwdirks:
> I have a style question on these changes. Since the 'getters' and 'setters' are attributes, they don't need to be invoked as functions anymore. Maybe the references should something like this:
> 
> ```
> s = self.port.dsr
> self.port.rts = False
> ```

I don't believe it's merely the pySerial client's stylistic choice - using it with and without the parenthesis represent different operations.  In one case you're directly reading/write a value from an objects "member values"; in the other case you're calling a function defined by the object to return the value to you.  The latter is far more flexible to the object creator - you only need to have the value be valid when the 'get' function is called.

Sure enough, though, it looks like the pySerial package is set up to use the direct access style you'd favor:

    def setDTR(self, value=1):
        self.dtr = value

which is to say, calling the old-style "setDTR(True)" function is simply changing the value of 'dtr' as passed.

I've updated the change to read as you suggested, which looks like the ONLY correct way to use the new PySerial interfaces.

leskerr:
This works fine. Thank you. Merge at will!

AESilky:
I haven't taken the time to look at the PySerial code, but is it possible that they used `@property` to set up the member value with a getter/setter that are used when you access it? The `@property` decorator was added in Python3. With that, `port.dsr` might not be directly reading/writing the value, it might be going through get/set methods.

AESilky:
Python3 added the decorator to make python code easier to read/write while still allowing classes to encapsulate class attributes.

pwdirks:
Not even something I was aware of.  Looks like a member value reference, turns out to be a function call... Hmmm, not sure I'm fond of the style but it certainly works!

AESilky:
They did it because one of Python's tenets is to make code easy to write
and clear to read. At the same time, they wanted to support encapsulation.

The hope is that by using this feature it makes the client code easy and
clear and it's fine because the client doesn't need to know if processing
is being done or not when the value is accessed.

C# has always had this same capability (easily add set/get code to member
value access). In C# code you shouldn't have any 'get'/'set' methods - most
style checkers have a rule to flag these methods as it clutters the client
code with unnecessary '()' pairs.

PR #126, opened 2020-08-13
Es 104 record playback

AESilky:
From the call today, I'm going to merge this into master. There are a few known things I'm working on, but the core functionality seems good.

Working on:
1. MKOB is currently hardcoded to record its session. This should be a GUI option to turn recording on/off.
2. MKOB recording filename is hardcoded to 'MKOB.timestamp.json'. You should be able to set the filename when you enable recording.
3. MKOB should add '[R]' to the station name when recording is enabled so others know the session is being recorded (but not save that as the configured station name).
4. The 'condense dead air' time is currently hardcoded to 5 seconds. This should be a parameter in the recorder class so different times can be used during playback (including not condensing).


pwdirks:
> #3 MKOB should add '[R]' to the station name when recording is enabled so others know the session is being recorded (but not save that as the configured station name).

This one actually surprised me - I always figured recording would be on/off on at least a full wire's granularity but it might be interesting to record just one station, or a few stations, and not others on a wire.  We might want to give some thought how to do that without overwhelming the UI.

Do you plan to also offer the ability to record a whole wire, including all current and future joiners?

leskerr:
@AESilky I had fun playing back your session on wire 11. My first impression was that the Morse was very choppy and hard to read. There was extra space between letters, and even sometimes extra space within an internally spaced letter.

I took a look at the `Recorder` class and tried deleting the `time.sleep(timediff)` statement. (Actually you have to replace it with `pass` or remove the preceding 'if' statement for it to parse without an error.) Once I did that, the Morse sounded normal. I didn't look at the code carefully enough to figure out what the proper fix might be, but this is definitely a clue.

I really enjoyed listening to the code and watching the code sequences at the same time. When I got to a point I wanted to analyze more carefully, such as when Derek turned the wire over to you, I did a Ctrl-C to freeze the display. I could confirm that Derek's 'latching' code packet at the end of his transmission didn't get lost in the network.

By the way, I also copied and pasted a portion of the session into a separate JSON file so I could listen to that segment over and over.

I'm looking forward to recording my next session on wire 11. This is already a very useful tool, even in its current form.

leskerr:
I like your idea of the [R] flag so other users know they're being recorded. I was already anticipating that privacy issue but hadn't thought of a way of dealing with it. I'm going to add the [R] flag to my station ID manually until MKOB has a provision for doing it automatically.

AESilky:
@pwderks - The recorder records all of the traffic (not sophisticated at all). Adding the '[R]' is just to indicate that the wire is being recorded.

AESilky:
@leskerr - I really appreciate your feedback on the timing - your ear is far (far, far, ...) superior to mine! I put in the `sleep` in order to maintain the overall pace of the transmission. Without that, a pause of 2 or 3 seconds while someone was thinking about what to send next would be lost. However, it is very possible that my simplistic approach isn't good enough. I probably need to account for how long it takes to play (sound) the code sequence before reading the next line.

AESilky:
@leskerr - your comment about cutting out a portion of the file and then being able to just play that portion helps with my justification of the 'each line contains all the info' format.

leskerr:
> I put in the `sleep` in order to maintain the overall pace of the transmission. Without that, a pause of 2 or 3 seconds while someone was thinking about what to send next would be lost. However, it is very possible that my simplistic approach isn't good enough. I probably need to account for how long it takes to play (sound) the code sequence before reading the next line.

I haven't thought this through thoroughly, but I don't think `Recorder.playback` needs its own timer to pace itself line by line through the file. `KOB.sounder` doesn't return until it has finished sounding out the code sequence, and when it's called with the next code sequence it waits the appropriate length of time before processing that character.

This should work for pauses up to three seconds or so (as I recall). Anything longer than that, `KOB.sounder` reduces the pause to about half a second (if memory serves).

AESilky:
I'll do some testing with clock set to different intervals without
condensing dead air (wire) and see what it takes to get the playback timed
correctly.

AESilky:
Testing with Clock set to announce every three minutes - if I want a true playback I need the sleep. Otherwise all of the three minute gaps are lost.

I understand that we probably don't want the sleep for the short gaps.

I plan to record wire-11 in both morse and (actual) audio to be able to compare them and be able to make them match up.

AESilky:
Recordings of Wire-11 Saturday 7/15. Audio, MKOB, and KOB
[MKOB-Wire11-20200715.zip](https://github.com/MorseKOB/PyKOB/files/5079337/MKOB-Wire11-20200715.zip)
(Audio) https://1drv.ms/u/s!AqnBbsz_ABrPhZU0BndZCzqX6zWsvQ

PR #132, opened 2020-09-03
Lk 130 restructure local loop

leskerr:


MKOB/kobactions.py
```diff
@@ -65,12 +65,15 @@ def doHelpAbout():
...
+def doOfficeID(event):
     kc.OfficeID = kw.varOfficeID.get()
     config.set_station(kc.OfficeID)
     config.save_config()
     km.myInternet.set_officeID(kc.OfficeID)
 
+def doCircuitCloser():
+    km.from_circuit_closer(kw.varCircuitCloser.get() == 1)
+
 def doWPM(event=None):
```
AESilky, line 77: Just curious...
Why does one have `event`, one doesn't have `event`, and one has `event=None` and none of them appear to use `event`? Is that part of the API for the GUI (I'm not too familiar with it).

MKOB/kobmain.py
```diff
@@ -43,60 +43,92 @@
...
 
-kob_latched = True
-keyboard_latched = True
-internet_latched = True
-latch_code = (-1000, +1)  # code sequence to force latching
+local_loop_active = False  # True if sending on key or keyboard
+internet_active = False  # True if a remote station is sending
+
+latch_code = (-0x7fff, +1)  # code sequence to force latching
+unlatch_code = (-0x7fff, +2)  # code sequence to unlatch
```
AESilky, line 50: I like it!

MKOB/kobmain.py
```diff
@@ -175,23 +208,16 @@ def readerCallback(char, spacing):
...
-    """log the current latching states and reinitialize"""
-    global kob_latched, keyboard_latched, internet_latched
-    print("Reset\n kob_latched: {}\n keyboard_latched: {}\n internet_latched: {}"
-            .format(kob_latched, keyboard_latched, internet_latched))
-    kob_latched = True
-    keyboard_latched = True
-    internet_latched = True
+    """log the current internet state and regain control of the wire"""
+    global internet_active
+    print(
```
AESilky, line 213: Would you want to use `log.info(...)` here. We kind of got away from using the Log, but we should probably go back to it. We can add functionality to 'Log' over time to be able to log to a file, to the console, to both, etc. It currently adds a timestamp, which can come in handy.

MKOB/kobactions.py
leskerr: Some of the widgets pass an event parameter to their respective handlers and some don't. And some of the handlers are called directly from within the MKOB modules, and those calls never pass an event. I could have kept things simple and specified `event=None` for all the handlers, because that always works.

MKOB/kobmain.py
leskerr: I'm on the verge of deleting that print statement and not logging the information.

keytest.py
```diff
@@ -0,0 +1,6 @@
+from pykob import kob
+
+myKOB = kob.KOB(port="com3")
+myKOB.setSounder(False)
+while True:
+    print(myKOB.key())
```
AESilky, line 6: Can we do like I did in 'Closer-Test.py' and use the 'config' module to get the port?

```
Closer-Test.py:
import sys
from pykob import kob
from pykob import config

try:
    interface = config.serial_port

    myKOB = kob.KOB(port=interface)
    myKOB.setSounder(False)
    while True:
        print(myKOB.key())
except KeyboardInterrupt:
    print()
    print("Thank you for using the Closer-Test!")
    sys.exit(0)     # Indicate this was a normal exit
    
```
Looking at it, they seem to be doing the same thing. I checked 'Closer-Test.py' in from the code I got from you when you asked if I could test on my Mac and Linux machines.

AESilky (approved):
I put in a few questions/comments, but nothing to keep me from approving.

keytest.py
AESilky: I did this because the Mac and Linux don't have 'com3' and my Windows machine uses 'com5'. I also wanted to break out without getting a back-trace.

MKOB/kobactions.py
AESilky: Makes sense (I kind of assumed that). Might be a good change to keep the question from coming up in the future (since it's pretty common for event handlers to get an event passed in). Maybe when this file gets changed in the future.

MKOB/kobmain.py
AESilky: If you want to that's fine. 

At some point I expect we will incorporate a logging package. They generally provide a lot of control about what is and isn't output to the log (without changing any code) and most even allow you to change the verbose-ness dynamically while the app is running. Since we don't have that in place yet, if you don't want to clutter the output you can remove it. It didn't seem to occur too often, so it didn't bother me, though I can see it being better as `log.debug(...)`.

leskerr:
I'm not sure if I ever replied to this. I removed `keytest.py` from the MKOB project. It was a one-shot throwaway script that was never intended to be kept.

AESilky:
Sure. I added a version that takes into account 'config'. We can delete it at the point it isn't useful.

PR #139, opened 2020-09-15
Improve playback fidelity (from @leskerr)

AESilky:
Use `KOB.sounder` to handle pauses less than 1 sec (which it does reliably). `Recorder.playback` handles longer pauses.

### DIFF
--- a/pykob/kob.py
+++ b/pykob/kob.py
@@ -104,15 +104,15 @@ class KOB:
     def key(self):
         code = ()
         while True:
+            t = time.time()
             s = self.port.dsr
             if s != self.keyState:
                 self.keyState = s
-                t = time.time()
                 dt = int((t - self.tLastKey) * 1000)
                 self.tLastKey = t
                 if self.echo:
                     self.setSounder(s)
-                time.sleep(DEBOUNCE)  # MAYBE COMPUTE THIS BASED ON CURRENT TIME
+                time.sleep(DEBOUNCE)
                 if s:
                     code += (-dt,)
                 elif self.cktClose:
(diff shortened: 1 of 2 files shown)


AESilky:
With the latest changes I get this error when playing your 'one-two-three.json':

```
esilky@Eds-Macbook PyKOB % python3 Capture-Replay/play.py ~/code/_temp/OneDrive-2020-09-07/one-two-three.json
2020-09-14 23:20:43 	INFO: 	Interface for key and sounder on serial port '/dev/tty.usbserial-AO007TCT' not available. Key and sounder will not be used.
KOB.sounder buffer empty: 1.006 -0.003 2 [-32767, 2]
  File "Capture-Replay/play.py", line 63, in <module>
    myRecorder.playback(myKOB, showDateTime=args.showDatetime)
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/pykob/recorder.py", line 133, in playback
    pause = -code[0] / 1000.0  # delay since end of previous code sequence and beginning of this one
IndexError: list index out of range
```

leskerr:
@AESilky - Oops, I anticipated that situation but never tested it. I mistakenly had `next` instead of `continue` in my code. No syntax error, and `next` has no effect at run-time.

I fixed the mistake, checked that it works, and pushed the commit to the server.

'two' and 'three' still run together on playback, but at this point I don't feel like doing anything about it. My overriding priority is fidelity of playback for hand-sent Morse.

AESilky:
That makes sense. Could it have anything to do with the length of the audio recordings for dit and dah?

PR #149, opened 2020-09-18
Add interface type config parameter for use by pykob.kob

### LINKED ISSUE #147: Add config option, Interface, with values of 'Key & Sounder', 'Loop', and 'Keyer'
The kob code needs to know whether a separate key & sounder, a KOB using a loop interface, or a keyer is being used in order to drive the interface and detect and process the inputs correctly.

To do that a new configuration option needs to be added to allow the interface type to be specified.

Config property name:
`interface`
Valid values (enum):
- `key_sounder` (1)
- `loop` (2)
- `keyer` (3)

AESilky:
Added the config value to `config.py` and the ability to set it and list it to `Configure.py`. Put a placeholder parameter in `kob.py`.

Closes #147

### DIFF
--- a/Configure.py
+++ b/Configure.py
@@ -33,6 +33,7 @@ Command line parameters:
     -p | --port <port>                 : Set the system PORT (COM or tty) value to use for a sounder
     -L | --local <ON|OFF>              : Enable/disable local copy of code  
     -R | --remote <ON|OFF>             : Enable/disable sending code to remote wire (over the internet)
+    -I | --interface KEY_SOUNDER|LOOP|KEYER : Set the interface type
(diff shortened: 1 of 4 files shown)


leskerr (commented):
The parameter `echo=False` can be deleted from the line
```
def __init__(self, port=None, interfaceType=config.interface_type.loop, audio=False, echo=False, callback=None):
```
in  `pykob/kob.py`, since that parameter is superseded by the new `interfaceType` parameter.

The two lines that read
```
if self.echo:
```
can be changed to
```
if self.interfaceType == config.interface_type.key_sounder:
```
And

```
self.interfaceType = interface_type
```
needs to be added to the `__init__` function.

AESilky:
Thanks for the review. I'll make those changes and merge.

PR #151, opened 2020-09-24
Lk 111 implement 'key & sounder' interface

leskerr:
I ran a successful test using the Key & Sounder interface on wire 11.


AESilky (approved):
Looks good.

PR #153, opened 2020-10-03
Add speed factor and max silence to recorder.play and CLI options to …

### LINKED ISSUE #152: Provide 'speed_factor' parameter to recorder.play to allow playback speed adjustment
When playing back recordings sometimes the speed of the code is too fast for the person that is trying to copy it. It would be helpful if a `speed_factor` value could be passed in. The value would be a percentage of the recorded speed that the code should be played back at. So a value of '100' would play back at 100% (the same speed it was recorded at). A value of 66 would play back at 2/3rds the speed the code was recorded at.

AESilky:
…play.py.

In `recorder.play` the speed_factor is a percentage of the actual recorded speed. The max_silence is the maximum amount of silence (dead air) that should be played.

The `play.py` command provides --speedfactor and --maxsilence CLI options to specify the values.

Closes #152


leskerr (commented):
Works fine. Thank you!

Merge at will.

AESilky:
Did you try the max silence as well? I used 'maxsilence 2' and played the recording from the modem high-speed session with the --list option so I could monitor the timestamp and see when the silence was reduced to 2 seconds (it occurs a couple times after a couple minutes).

leskerr:
I tried reducing max silence to 3 seconds. It worked fine. I forgot to try the --list option.

PR #154, opened 2020-10-04
Add play to file menu

AESilky:


AESilky:
I notice that there are some changes that were made in master that went back in this code. For example, I notice that in kobmain the 'mySender', 'myReader', 'myRecorder' and 'myInternet' changed back to 'Sender', 'Reader', 'Recorder' and 'Internet'. I'd need to go through history and see when those changed to try to get them back during a merge.

leskerr (commented):
It's fine with me if you'd like to merge these changes into master.

I tried replaying Monday evening's wire 11 session. I watched the reader window carefully and it never jumped. I only had time (i.e. patience) to watch the first half hour, and as I recall the session went for an hour or so. So I may have missed the instance of the window jumping if it occurred during the second half of the session.

What I'm going to do is pay attention during live sessions, and if the window jumps I'll note the time when it occurred. That way I can play it back later and know when to watch for it without having to pay attention to the whole thing.

As for the station list, I noticed that when a new station started sending, the station list cleared and only listed the current station, not all the others that were listening. At the beginning it showed all the stations that sent at some point during the session, which is nice.

AESilky:
I will look at the station list again. I must not have played past a sender change after implementing that part.

Issue #57, opened 2020-05-27
MKOB: Reader window jitters when full


leskerr:
When the reader window is full, new text should cause the window to scroll up smoothly. Instead, it jitters up and down slightly with each new character.

This problem was frustrating enough that it caused me to give up on the whole MKOB project in despair.

AESilky:
As Tim Allen said in GalaxyQuest - "Never give up. Never surrender!"
https://www.youtube.com/watch?v=9fdcIwHKd_s

leskerr:
I followed Tim Allen's advice and took another crack at my old nemesis, the longstanding jitter problem in the code reader window. Yesterday I had a breakthrough and today I can happily announce that I've found a solution. I'm switching from Tkinter's Text and Scrollbar widgets to Tkinter.ttk's ScrolledText widget, and the problem has gone away. Yay!!

I'm experimenting with using Tkinter.ttk across the board for the MorseKOB GUI. Supposedly it provides a more modern and more native look and feel across platforms (Windows, macOS, and Linux) than Tkinter. It'll take me a few days to wade through all of this, so please be patient. I'm not as fast as I used to be, and I never was very fast to begin with. When I'm done, though, I'll create a pull request and ask you guys to kick the tires on the new interface and send me Mac and Linux screenshots.

In the meantime, please join me in celebrating.

AESilky:
That's fantastic! Can't wait to see/try it.

AESilky:
I pulled the branch `lk-57-Fix-reader-window-jitter` and tested. Here are screenshots from Mac with some text in the Reader section before it is full and then once it has become full and scrolled.

<img width="912" alt="Screen Shot 2020-06-06 at 11 42 02 PM" src="https://user-images.githubusercontent.com/19332027/83962184-2633cb00-a850-11ea-875e-9d1a28950461.png">
<img width="912" alt="Screen Shot 2020-06-06 at 11 44 07 PM" src="https://user-images.githubusercontent.com/19332027/83962189-2e8c0600-a850-11ea-9106-db7875b8f90f.png">

The scroll in the section is smooth.

leskerr:
Thanks for the screenshots, @AESilky. It's good to know it's working on the Mac.

This version just addresses the jitter in the reader window. I'm going to do another commit on the same branch that uses Tkinter.ttk for all the widgets and see if it gives a better looking interface. If it doesn't work out, I assume there's some way to roll back to the present commit and throw away the later one.

[...]

PR #155, opened 2020-10-06
Implement code sender checkboxes

leskerr:
Implement `On` and `Repeat` controls for the keyboard.

Also, change the code reader window scrolling method from `yview_moveto` to `see`. If we're lucky, this will eliminate the jumping of the code reader window (issue #57).


AESilky (approved):
Looks good.

PR #164, opened 2020-10-22
Es 150 modem support

AESilky:
Add a config option to invert the key input. This is needed because the input from a modem is opposite from the key/loop interface. This option is used to allow input from a modem being used with the dial-up morse (or recordings of dial-up morse sessions).

While using the modem and starting MKOB while the modem was already sending code I found that it (quite often) generated a 'None method access' error. I tracked this down to the the fact that the kobmain code (outside of the method definitions) was being executed when kobmain is imported by kobactions (which is imported by kobwindow, which in turn is imported by MKOB.pyw). This happens before the GUI is created and initialized. The kobmain code was creating a kob object that included callbacks to update the reader and sender windows, but the windows do not yet exist at that point.

I moved the kobmain code into an `init` method that kobwindow calls once everything is created and initialized. I tested by starting MKOB multiple times with the modem already sending code and no longer got a 'None method access' error.


leskerr:
For modem compatibility, you should also invert the sounder output (RTS).
Maybe you're doing this already.

jchausler:
If we are considering supporting modems for connection to a dial-up system I ran into another problem when I tried to do this with an auto answer modem.  If MorseKOB 2.5 was connected to a wire that was active and thus the transmit data line into the modem was active, the modem wouldn't auto answer (original Hayes 300).  Some way to monitor the modem's CTS line and not key the transmit data unless the CTS line says there's a valid modem connection.  If the transmit data line was not active the modem would answer an incoming call from a dial-up system just fine.  Just FYI...73, Chris Hausler-----Original Message-----

AESilky:
@leskerr So that would correctly send from the keyboard? I know how I have it hooked up for transferring the recordings to our digital format. I'm not clear how I would set it all up if I actually wanted to use it with a dial-up hub.

AESilky:
@leskerr and @jchausler It seems like to be able to use MKOB with a modem to connect to a dial-up hub it would need to use 2 serial ports. One to connect to a key & sounder (as it does now) and one to connect to the modem to use instead of the network to connect to the dial-up hub. That port would need the DTR and the RTS inverted to work correctly with the modem.

I'm I just not thinking correctly about that?

If that is true, that wasn't my intention when creating this. I was just trying to support the setup needed to transfer modem recordings.

That setup connects the modem received data to the DTR (key input) into MKOB to receive code data from the modem. It connects MKOB sounder out (RTS) to the sounder interface to allow monitoring the code as it is processed. In this case, I don't think the RTS needs to be inverted, since it is MKOB sending the code to the sounder and it was already inverted as it was being received.

I'm I not thinking about things correctly?

jchausler:
Hi Ed,Actually, no.  In the case of the modem function, the modem link is essentially the key and sounder.  What it allows is a way to connect a dial-up system to the MorseKOB network, the dial-up connection essentially becoming the key and sounder.  You could still monitor what was happening on that instance of MorseKOB by turning on sound.  However, what I did was bring up a separate copy of MorseKOB on my computer which did have a key and sounder on it.  Since I did not turn on the sound option on the MorseKOB instance connected to the modem, nothing was heard on that connection, I just heard what was coming over that link after the usual short delay on the copy running with the key and sounder.  I could also then send and it would be sent out the dial-up line.  Without going into details what I was doing with this was dialing into my auto-answer Hayes 300 from a remote location with a regular dial-up set and with MorseKOB connected to one of its wires I could communicate with others on that wire using a dial-up set at a location other than where my computer and Internet connection was located.  The only problem I had doing this was if there was activity on the MorseKOB wire when I dialed in, the Hayes would not auto answer because of the activity on its transmit data input.  That was the reason for wanting something within MorseKOB in modem mode to monitor the CTS signal from the modem and not send anything to the transmit data line unless CTS was true.  That would have fixed the problem.  What I did to get around this was build what essentially was a sounder driver circuit to drive a Morse relay with separate front and back contacts from the modem's CTS signal with the common of the relay contacts connected to the modem's transmit data line and the front contact connected to the transmitted data from MorseKOB.  The back contact was connected to a negative voltage to hold the modems transmit data line low, unless its CTS was true indicating that there was a connection, so it would auto answer.  Thus with CTS false this also kept the receive data line from the modem in a marking or "key closed" state just like a real telegraph office with its circuit closer closed to keep the circuit complete for whomever was sending on MorseKOB.  Another option which I played with at least was to be with my computer and dial out with the Hayes to the Canadian hub.  If I got a decent connection with the hub (which unfortunately didn't happen often) anyone dialed into the hub could communicate over MorseKOB through my computer.  My ID string indicated that it was the hub as all the different folks on the hub just showed up as this one ID on MorseKOB.  But I only played with this.  I was never successful making such a connection on Morse Day and in fact found few or no folks on the Canadian hub on the several Morse Days I tried to connect.  Anyway, that's the uses for the modem option that I'm talking about.73, Chris-----Original Message-----

leskerr:
> While using the modem and starting MKOB while the modem was already sending code I found that it (quite often) generated a 'None method access' error. I tracked this down to the the fact that the kobmain code (outside of the method definitions) was being executed when kobmain is imported by kobactions (which is imported by kobwindow, which in turn is imported by MKOB.pyw). This happens before the GUI is created and initialized. The kobmain code was creating a kob object that included callbacks to update the reader and sender windows, but the windows do not yet exist at that point.
> 
> I moved the kobmain code into an `init` method that kobwindow calls once everything is created and initialized. I tested by starting MKOB multiple times with the modem already sending code and no longer got a 'None method access' error.

That sounds like a very good change. As you said, it might clear up some other problems I was running into on startup.

leskerr:
> @leskerr So that would correctly send from the keyboard? I know how I have it hooked up for transferring the recordings to our digital format. I'm not clear how I would set it all up if I actually wanted to use it with a dial-up hub.

Yes, if you want to send on the keyboard to the modem, then the RTS line needs to be inverted as well. Not that we have dial-up Morse users anymore, as far as I know.

leskerr:
> @leskerr and @jchausler It seems like to be able to use MKOB with a modem to connect to a dial-up hub it would need to use 2 serial ports. One to connect to a key & sounder (as it does now) and one to connect to the modem to use instead of the network to connect to the dial-up hub. That port would need the DTR and the RTS inverted to work correctly with the modem.

I used MorseKOB 2.5 with the dial-up hub in one of two ways. One way was to use the keyboard to send. The other way was to connect to one of the KOB wires (typically wire 11, no surprise there), thereby establishing a gateway between dial-up users and MorseKOB users. It sounds like @jchausler was doing something similar. The gateway was a novelty but never caught on. When the US hub went away, the gateway became a thing of the past.

Historically we've always had a hard time connecting to the Canadian dial-up hub from US phone lines. I tried very hard to persuade the operator of the Canadian hub to install a MorseKOB computer at the Canadian hub site. It would have been a great solution and easy to implement, but for whatever reason it was never acted on.

And just to avoid possible future confusion, the key input is the DSR line, not DTR. (@AESilky you know that already. :-)

leskerr:
@AESilky I think you should leave things the way they are, and only have the option for inverting the key input.

If we ever do decide to implement a full-blown modem option, though, there's one more thing to consider in addition to the CTS line issue that @jchausler described, and that has to do with noise filtering.

When the key input is coming from a physical key, then the program needs to deal with contact bounce. Otherwise, when the key is closed, the program sees a burst of very short transitions before the input settles down. These contact bounce transitions can last for 5 ms, more or less.

PyKOB handles this by responding to the first transition and then ignoring subsequent transitions for 10 ms. This works fine for a physical key, but not so well for a modem connection.

When connecting to the dial-up hub over a phone line, I found that every now and then there'd be a brief noise blip. It might last only a few milliseconds, but the program would see it as being a very short dot. Interestingly, such a short dot can't be heard on a real sounder, because the mechanical nature of the sounder doesn't respond to such a short noise pulse. But the debounce algorithm extends that brief pulse to a 10 ms dot. This can clearly be heard on the simulated sounder, and it also interferes with the decoding algorithm.

Because of this, MorseKOB 2.5 uses a different approach when the `modem` mode is enabled. It doesn't respond at first when it receives a transition, but waits 10 ms to see if the state of the input has really changed. If it has, then it processes the transition appropriately. If not, it disregards what it assumes was a noise pulse.

But that's not quite the end of the story. <sigh> It turns out MorseKOB 2.5 has a bug. When I added the `loop` mode (which came into being after the `modem` mode), I accidentally used the noise suppression algorithm instead of the debounce algorithm. It usually doesn't make any difference, except in the rare case where someone sends a very short dot. The sending operator doesn't notice, because their sounder responds directly to their key (they're in the same loop, after all). But the program thinks the short dot is noise and filters it out, so anyone else on the wire doesn't hear the short dot at all. I became aware of this bug long after I lost the ability to change MorseKOB 2.5.

AESilky:
This is great information. I'm glad that we've recorded it.

I agree @leskerr . I think it best to leave the change as-is. I labeled the config option 'Invert Key Input', which is accurate. As I mentioned, the setup to transfer the modem recordings to MKOB recordings works correctly with just the key inverted (if I have it on, the sounder is correct).

I will however, change the name of the issue to reflect the actual change. That way, if anyone reviews issues they won't think we added dial-up support.

leskerr:
Perfect!

PR #165, opened 2020-10-23
Playback - respect time delay between station changes.

AESilky:
When a station change occurred the true (timestamp) delay wasn't being taken into account. This caused 'kob' to do a minimal delay which didn't reflect the actual time delay between the change of senders.

Test for station change and if one occurred do a delay based on the timestamp to be accurate.


AESilky:
I tested with a couple wire-11 sessions. One copy that I reduced to a few seconds of each sender so I could test a dozen sender changes within a minute or so. The time between senders in the playback resembles what I recall from the session.

Issue #159, opened 2020-10-18
Fix buffering


leskerr:
The program needs to buffer the received Morse to smooth out the variation in the length of Morse characters as well as any jitter in the internet (see https://github.com/MorseKOB/PyKOB/wiki/Morse-playback#buffering). There are also situations where buffering needs special handling, such as when an operator first opens their key.

PR #170, opened 2020-11-04
Reorg project: move MKOB files up to the root level.

AESilky:
Discussed with other developers and this makes things easier for development and deployment.


AESilky:
Since this is a change in the file structure, I would like to do this while there aren't too many branches open. For the branches that do exist, it is probably best to delete them and then recreate them from the new structure (unless they are only modifying pykob or base level (CLI) content).

leskerr (approved):
I have no branches open at this time. In fact, I've been waiting for this restructuring before working on issue #159.

Is now the time to remove `setup-mkob.py`?

PR #172, opened 2020-11-06
MKOB Station List update converted to being message driven.

AESilky:
Updated the Station-List to use messages to update the contents. Tested with a Wire-11 recording to check that stations and senders displayed correctly.

Will update the Reader window next.

I don't want to tie this to closing the issue, as the Reader window also needs to be converted to using messages. Then, I would like some testing before closing the issue.

### DIFF
--- a/kobactions.py
+++ b/kobactions.py
@@ -35,6 +35,7 @@ import time
 from pykob import config, kob, internet, morse, recorder
 import kobconfig as kc
 import kobmain as km
+import kobreader as krdr
 import kobstationlist as ks
 
 import pykob  # for version number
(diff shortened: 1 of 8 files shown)


leskerr (commented):
I tried this version on wire 11 today. There was no sign of flicker, jitter, jumping, or duplication in the reader and station list windows. That's definitely a good sign, and I'm optimistic that the code is now on a much more solid footing.

I did notice some problems. They may all be different symptoms of the same thing:

- When I was sending, my line of the station list didn't get moved to the bottom of the list.
- When a new station started sending, their identification didn't show up in the reader window.
- On playback, none of the other stations show up in the station list.
- The playback controls for skipping to the previous sender or next sender don't work.

Here's a clue: The station IDs in the capture file always show my own ID, no matter who's sending.

@AESilky I'm happy if you want to merge this in with master, and deal with the above problems as a new issue or a continuation of the same issue. If you'd rather not merge now, what's the best way for me to proceed with a new branch? Should I create a branch off the current master, or should I base my new branch on this one?

AESilky:
I probably missed a place where incoming senders were changing. I'll take a look at that tonight.

I'm confident that this is the approach we should use to update the windows/controls given the multiple threads needed to perform the functions. I'll go ahead and merge the change so you can pull master and create a branch from it.

AESilky:
The fact that the recording file only has your station probably explains why the playback keys don't work. But that's a good test, as it should probably still go to the beginning and end of the recording (given that the playback only knows of one sender).

Thanks for providing all of the info. I'll create issues for those so I can check them off as I do them. Hopefully I can get them done by Monday evening so I can join Wire11 with it and test it out.

Issue #166, opened 2020-10-29
MKOB sometimes locks up when disconnecting from wire


leskerr:
Several times lately the program has become unresponsive while I've been using it. When this happens, the only recourse is to force a hard shutdown. This is with Windows, and I believe it has always happened when I've tried to disconnect from the wire while a transmission is in progress.

I've tried unsuccessfully to force it to happen by repeatedly connecting and disconnecting. If it happens again, I'll log the occurrence here. I encourage others to do the same, so we track how persistent the problem is.

AESilky:
I was trying to test/force this by repeatedly connecting/disconnecting from Wire-107. After ~10 connect/disconnect actions the GUI locked up. I will work with this process and try to track down the cause.

AESilky:
I think the problem might be caused by multiple threads updating the GUI. There was an issue reported for the Station List occasionally duplicating. I put in a lock (RLock) to protect multiple updates - but that was (probably) neive.

The tkinter GUI framework (like the majority of GUI libraries) is 'single-threaded'. What that means is that it only supports GUI interface operations from one thread at a time. Since it's not really known when the Main GUI thread might update a UI component, the general pattern is to make sure that only the Main GUI thread ever updates the GUI components.

In debugging, I've found that a number of threads are updating the GUI components. The most common component to be updated by different threads is the Station List. There are other GUI components that are updated by different threads, but they happen much less frequently. That only helps the problem because it is much less likely to happen.

I believe the fix to this is to make sure that only the Main GUI thread makes changes to the GUI components.

leskerr:
Thanks for figuring this out! I wish I had known about this from the beginning. I hope it's not too difficult to restructure things. I'd do it myself if I knew how. Maybe it's a good time for me to learn.

AESilky:
Well, we'll see...

I haven't done tons of GUI work, but when I have (WPF, WinForms, Swing, Autodesk-ACAD and Revit) I have been hit by the single-thread update restriction while trying to run multiple threads for operations within the application.

First, I'll work on the Station List - as that is the GUI component that is most frequently updated by different threads. The lock I added is probably what is causing the lockups. However, without the lock we get the occasional (distracting) list duplication.
I will try to make this work well with a pattern to allow the main thread to perform the update from data provided by the internet or recorder-playback thread.

AESilky:
Found this and it goes along with the guidelines I've heard (and used) for quite a few years with other GUI frameworks.
[...]

Issue #177, opened 2020-11-09
MKOB: Changing wire connected to doesn't clear the Senders window


AESilky:
When testing the branch to fix #166 that changed windows updates to be message based;
@AESilky  noted: After playing a recording (which populated the Senders window) I connected to Wire 105. The Sender window didn't clear; the 'Beaufort Scale, 30 wpm, AC' was added, but the previous list of senders (from the recording) remained.

Proper functionality is to clear the Senders window when there is a connection to a new wire.

AESilky:
I don't have time to fix it now, but I believe that the problem is that the 'Recorder' station-list thread is still running after the recording has finished. That causes the Station window to update with the stations that were part of the recording.

The 'Recorder' station-list thread needs to stop when the recording is done/stopped.

PR #178, opened 2020-11-12
Es 166 lockups

### LINKED ISSUE #173: MKOB: Latest sender doesn't move to the bottom of the Senders List
When testing the branch to fix #166 that changed windows updates to be message based;
@leskerr noted: When I was sending, my line of the station list didn't get moved to the bottom of the list.

Proper operation is that:
1. The current sender is listed at the bottom of the list
2. Next up the list should be the previous sender
3. The list should continue to move up in the order a station last sent
4. Stations that are connected but haven't sent, should be listed next up the list

AESilky:
This intends to solve a number of issues.

Primarily #166. Also fixes #173 , #177 
It does seem to have introduced some problems with playback, but it would be nice to get realtime function working well and then adjust playback as needed.


leskerr (commented):
I tried this version with two instances of MorseKOB 2.5 running at the same time (on the same computer) to simulate a three-way session. Generally it worked well, but I did notice these problems:

- When I send, my ID doesn't move to the bottom of the station list.
- When I'm connected to a wire and I click on the Connect button, the program disconnects from the wire but the the red 'connect' indicator stays red and I can't reconnect.

Just for fun, I added a second instance of MKOB 4.0 to the mix. I was using my bug and sounder with one of the MKOB 4's, and different choices of simulated sounders with the other three programs (sending with the keyboard, of course). It worked fine, other than the problems noted above.

KOB x 4

AESilky:
For item 1, do you mean on the 4.0 you are sending from? Or on the other ones?
I will test with 4 instances on and off of a wire:
 - 2.5 and 4 on Win
 - 4 on Mac
 - 4 on Linux
and see what I get.

leskerr:
What I meant was when I'm sending on MKOB 4, my ID doesn't drop to the bottom of my station list.

AESilky:
Okay. Does it go to the bottom on the other MKOB 4 instance that is also connected?

leskerr:
Yes, it goes to the bottom of the station list on the other MKOB 4 instance, the one that's listening. It stays at the top of the MKOB 4 instance that's sending.

leskerr:
> * When I'm connected to a wire and I click on the Connect button, the program disconnects from the wire but the the red 'connect' indicator stays red and I can't reconnect.

Here's a clue. I used to have a one second delay in `kobmain.py` after disconnecting. I had it in there in an attempt to allow any pending incoming packets to be flushed, but that didn't seem to be successful. In my latest changes to get rid of unwanted delays, I commented out that delay. When I uncomment it, disconnecting works again. I still find the delay annoying, though.

```
        Internet.disconnect()
        Reader.flush()
        time.sleep(1.0)  # wait for any buffered code to complete
        connected = False  # just to make sure
```

AESilky:
Thanks for the hint. I'll try to debug the disconnect and see if I can tell where it's getting hung up.

AESilky:
I pushed a change that fixes the `connected` flag not being cleared on 'disconnect'. That got commented out along with the delay. 

I also changed the setting of the callbacks for the connected stations and the current sender to keep stations from showing up in the station list after it was cleared during disconnect. The 1 second delay was enough to keep this from happening, but without the delay stations would occasionally show up after the list was cleared. Now those callbacks are set on connect and cleared on disconnect.

leskerr (approved):
Well done! This is _much_ better than it ever was.

You can remove the commented out `time.sleep(1.0)` line if you like.

Just one minor problem remaining. If you disconnect from the wire while someone is sending, the local loop is left open. If the fix isn't straightforward, it can be left for another time. The workaround is to regain control of the local loop by toggling the Escape key.

AESilky:
I'll remove the sleep line and look quickly at that. If it's not obvious
I'll nether this in and open an issue for it. Then I can work it and the
local sender not moving to the bottom of the list.

AESilky:
@leskerr and I tested on Wire-11 today. Things worked well. Les noted a couple things, but I think it is fine to merge this now and then address the things he noticed.

PR #187, opened 2020-11-19
Change log.log to take one required parameter and update use of it wh…

AESilky:
…ere needed.

### DIFF
--- a/pykob/internet.py
+++ b/pykob/internet.py
@@ -117,14 +117,14 @@ class Internet:
                     self.rcvdSeqNo = seqNo
                     return code
             else:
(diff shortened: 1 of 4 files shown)


AESilky:
Log.log took two required parameters - a type and a message, but most use of it just passed a message. That was causing an error when trying to log the message saying that PyAudio wasn't available.

I changed it so the 'type' is a named parameter to allow all of the places that call it with a single message parameter to work. In the couple places that were passing a type I changed to use the other Log methods that add the type.

leskerr (approved):
@AESilky I like the way you're checking for whether Python is installed. Nice!

PR #193, opened 2020-11-28
Lk 192 improve morse decoding

leskerr:
These changes temporarily remove some experimental features from MKOB 4.0, making the program's behavior more predictable and more familiar to users who are accustomed to MorseKOB 2.5. These changes include the following:

- Disable automatic code speed recognition.
- Change ~ and + characters to underscores in the reader window.
- Don't display a long dash in the reader window to represent a long pause.
- Disable 'intelligent' spacing in the reader window.

These features may be reintroduced pending further development and testing.


AESilky (approved):
Changes are fine for now. 

Maybe we should have an option for the spacing and speed adjustment? I'll try this for a while, but I like the 4.0 behaviour compared to 2.5 when I run them side by side.

PR #198, opened 2020-12-04
Update version numbers

leskerr:
Also remove MKOB.cmd from PyKOB directory. A similar file exists in PyKOB/scripts.


AESilky:
I need to work on putting a release rule together that flags if version numbers haven't been updated when merging to master. I always forget to do it!

leskerr:
@AESilky I'm not sure the version numbers are that important in general, especially while we're pushing changes so frequently. I did it this time because I wanted to confirm that @jchausler and I were running the same version.

PR #199, opened 2020-12-06
Es 82 station list clear on wire change

### LINKED ISSUE #82: Station list sometimes fails when changing wire numbers
If you're receiving a feed on one wire and then switch to a different wire without disconnecting, the station list sometimes retains the ID of the previous feed.

AESilky:
For issue #82 there were two problems. First, if you had played a recording to the end, the station-list thread in the recorder-player kept running. This caused the stations from the recording to continue to be refreshed/added to the Station List. The second was that the connection to the wire was being changed with separate code from the Connect/Disconnect methods.

The changes are to:
1. Set the 'stop playback' flag when the playback completes normally (rather than being manually stopped).
2. Change the 'change wire' method to use the 'disconnect' and 'connect (toggle)' methods and also add clearing the 'sender ID' in the method.

This PR will CLOSE #82

### DIFF
--- a/kobmain.py
+++ b/kobmain.py
@@ -143,7 +143,9 @@ def toggle_connect():
     """connect or disconnect when user clicks on the Connect button"""
     global local_loop_active, internet_active
     global connected
+    global sender_ID
     if not connected:
+        sender_ID = ""
         ka.trigger_station_list_clear()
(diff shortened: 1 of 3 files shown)


AESilky:
@leskerr if this looks okay and works for you, you can go ahead and merge it and close the branch since I'm going to be busy with family during the day.

leskerr:
Looks good. Thank you!

Issue #197, opened 2020-12-04
Station list order sometimes incorrect


leskerr:
When a user is sending, their station ID should move to the bottom of their station list. The latest version of MKOB produces correct behavior for @leskerr and @AESilky, but not for @jchausler.

AESilky:
A few observations from a combination of @jchausler, @leskerr and @AESilky ...
1. It works correctly if only the GUI 'Circuit Closer' button or 'ESC' is used to open/close the circuit before sending
2. Typing in the keyboard window opens the circuit as long as the 'code sender' box is enabled (maybe not a bad feature, but not planned)
3. Opening and closing using the key's closer works, but doesn't move the local sender to the bottom of the list on the local machine
4. What to do if you try to close with 'Circuit Closer', '+', or ESC and the physical key closer is open?

For #4, I think the physical closer has to take priority over all of the rest. If a physical key is connected and enabled and the physical key is open, then the circuit is open, and nothing you do in the GUI/keyboard should change it. Only when the physical key-closer is closed (or a physical key isn't connected) should the GUI functions matter. @leskerr do you agree?

AESilky:
I am going to focus first on the physical key closer opening/closing and correctly moving the local sender to the bottom of the local list.

We can look at the other aspects once this is working as expected.

leskerr:
> 4. What to do if you try to close with 'Circuit Closer', '+', or ESC and the physical key closer is open?
> 
> For item 4, I think the physical closer has to take priority over all of the rest. If a physical key is connected and enabled and the physical key is open, then the circuit is open, and nothing you do in the GUI/keyboard should change it. Only when the physical key-closer is closed (or a physical key isn't connected) should the GUI functions matter. @leskerr do you agree?

Not necessarily. If the user's key lacks a circuit closer, then the Circuit Closer checkbox is the only way to close the circuit. In fact, this was the original intent of the checkbox. (The Escape key is a convenient alternative to clicking on the checkbox.)

AESilky:
@leskerr Got it. In fact, I know remember you telling me that before.

[...]

PR #203, opened 2020-12-09
Use posted messages to control Circuit Open/Close from Keyboard and Key

AESilky:
I made changes to trigger messages to update the open/close state from the various mechanisms that could cause it.
I did significant testing with 5 stations connected to a wire... 2 MorseKOB-2.5 and 3 MKOB4.

All of the connect/disconnect seems to work in various combinations. This also fixes the issue that if you just start typing in the keyboard window or just start keying on the key - it would send regardless of whether the circuit is open or not (now the circuit must be open for anything to be sent).

I tested with both a loop and a separate key-sounder.

However, I made some changes that I'm not sure how to test with the loop. I compared 4.0 master functionality to this change and also with 2.5. I honestly don't know whether I broke something or not - but I will say that none seem to do what I expect as far as local sound goes when I am sending with a key.

@leskerr - I think this fixes the main issue, plus the issues of sending when you simply type something or key something (whether the circuit is open or closed). I would like you to spend time testing the different scenarios. At the same time, I'm not sure about the 'loop' mode in conjunction with a 'synth' sounder. The 'synth' sounder didn't follow the key (but testing with 'master' it doesn't seem to either).

### DIFF
--- a/kobactions.py
+++ b/kobactions.py
@@ -126,6 +126,18 @@ def doConnect():
 #### Trigger event messages ###
 ####
 
+def trigger_circuit_close():
+    """
+    Generate an event to indicate that the circuit should be closed.
+    """
+    kw.root.event_generate(kobevents.EVENT_CIRCUIT_CLOSE, when='tail')
+
+def trigger_circuit_open():
+    """
+    Generate an event to indicate that the circuit should be opened.
+    """
+    kw.root.event_generate(kobevents.EVENT_CIRCUIT_OPEN, when='tail')
+
 def trigger_player_wire_change(id: int):
     """
     Generate an event to indicate that the wire number 
@@ -167,6 +179,18 @@ def trigger_update_station_active(id: str):
 ####
 #### Event (message) handlers
 ####
+
+def handle_circuit_close(event):
+    """
+    Close the circuit and trigger associated local functions (checkbox, etc.)
+    """
+    km.from_circuit_closer(True)
+
+def handle_circuit_open(event):
+    """
+    Open the circuit and trigger associated local functions (checkbox, sender, etc.)
+    """
+    km.from_circuit_closer(False)
 
 def handle_escape(event):
     """
--- a/kobevents.py
+++ b/kobevents.py
@@ -32,6 +32,8 @@ events.
 """
 
 # Define event messages
+EVENT_CIRCUIT_CLOSE = "<<Circuit_Close>>" # Close the local circuit
+EVENT_CIRCUIT_OPEN = "<<Circuit_Open>>" # Open the local circuit
 EVENT_CURRENT_SENDER = "<<Current_Sender>>" # Record the current sender (station)
 EVENT_PLAYER_WIRE_CHANGE = "<<Player_Wire_Change>>" # The player detected a wire change
 EVENT_READER_APPEND_TEXT = "<<Reader_Append_Text>>" # Append text to the reader window
--- a/kobkeyboard.py
+++ b/kobkeyboard.py
@@ -55,11 +55,12 @@ def keyboard_send():
             kw.txtKeyboard.tag_add('highlight', 'mark')
             c = kw.txtKeyboard.get('mark')
             if c == '~':
-                kw.varCircuitCloser.set(0)
-            code = km.Sender.encode(c)
-            km.from_keyboard(code)
-            if c == '+':
-                kw.varCircuitCloser.set(1)
+                ka.trigger_circuit_open()
+            elif c == '+':
+                ka.trigger_circuit_close()
+            else:
+                code = km.Sender.encode(c)
+                km.from_keyboard(code)
             kw.txtKeyboard.tag_remove('highlight', 'mark')
             kw.txtKeyboard.mark_set('mark', 'mark+1c')
         else:
--- a/kobmain.py
+++ b/kobmain.py
@@ -58,6 +58,10 @@ def __set_local_loop_active(state):
     """set local_loop_active state"""
     global local_loop_active
     local_loop_active = state
+    if local_loop_active:
+        if kc.config.interface_type == config.interface_type.loop:
+            KOB.setSounder(True)
+
 
 def __emit_code(code):
     """
@@ -94,8 +98,6 @@ def from_key(code):
             ka.trigger_circuit_open()
             return
     if not internet_active and local_loop_active:
-        if kc.config.interface_type == config.interface_type.loop:
-            KOB.setSounder(True)
         __emit_code(code)
 
 def from_keyboard(code):
@@ -153,10 +155,12 @@ def from_circuit_closer(state):
         Recorder.record(code, kob.CodeSource.local)
     if connected and kc.Remote:
         Internet.write(code)
-    if len(code) > 0 and code[-1] == +1:
+    if len(code) > 0 and code[-1] == 1:
+        # Unlatch
         __set_local_loop_active(False)
         Reader.flush()
     else:
+        # Latch
         __set_local_loop_active(True)
     ka.kw.varCircuitCloser.set(1 if not local_loop_active else 0)
 
--- a/kobmain.py
+++ b/kobmain.py
@@ -54,44 +54,62 @@ unlatch_code = (-0x7fff, +2)  # code sequence to unlatch
 
 sender_ID = ""
 
-def set_local_loop_active(state):
-    """set local_loop_active state and update Circuit Closer checkbox"""
+def __set_local_loop_active(state):
+    """set local_loop_active state"""
     global local_loop_active
     local_loop_active = state
-    ka.kw.varCircuitCloser.set(1 if not local_loop_active else 0)  # ZZZ is this GUI-safe? probably not
+
+def __emit_code(code):
+    """
+    Emit local code. That involves:
+    1. Record code if recording is enabled
+    2. Send code to the wire if connected
+
+    This is used from the key or the keyboard threads to emit code once they 
+    determine it should be emitted.
+    """
+    global connected
+    update_sender(kc.config.station)
+    Reader.decode(code)
+    Recorder.record(code, kob.CodeSource.local) # ZZZ ToDo: option to enable/disable recording
+    if connected and kc.Remote:
+        Internet.write(code)
 
 def from_key(code):
-    """handle inputs received from the external key"""
-    global internet_active
-    if not internet_active:
+    """
+    Handle inputs received from the external key.
+    Only send if the circuit is open.
+    Note: typically this will be the case, but it is possible to 
+     close the circuit from the GUI while the key's physical closer 
+     is still open.
+
+    Called from the 'KOB-KeyRead' thread.
+    """
+    global internet_active, local_loop_active
+    if len(code) > 0:
+        if code[-1] == 1:
+            ka.trigger_circuit_close()
+            return
+        elif code[-1] == 2:
+            ka.trigger_circuit_open()
+            return
+    if not internet_active and local_loop_active:
         if kc.config.interface_type == config.interface_type.loop:
             KOB.setSounder(True)
-        update_sender(kc.config.station)
-        Reader.decode(code)
-        Recorder.record(code, kob.CodeSource.local) # ZZZ ToDo: option to start/stop recording
-    if connected and kc.Remote:
-        Internet.write(code)
-    if len(code) > 0 and code[-1] == +1:
-        set_local_loop_active(False)
-    else:
-        set_local_loop_active(True)
+        __emit_code(code)
 
 def from_keyboard(code):
-    """handle inputs received from the keyboard sender"""
-    # ZZZ combine common code with `from_key()`
-    global internet_active
-    if not internet_active:
+    """
+    Handle inputs received from the keyboard sender.
+    Only send if the circuit is open.
+
+    Called from the 'Keyboard-Send' thread.
+    """
+    global internet_active, local_loop_active
+    if not internet_active and local_loop_active:
         if kc.Local:
             KOB.sounder(code)
-        update_sender(kc.config.station)
-        Reader.decode(code)
-        Recorder.record(code, kob.CodeSource.local)
-    if connected and kc.Remote:
-        Internet.write(code)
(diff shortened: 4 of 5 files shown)


AESilky:
This would CLOSE #197

leskerr (changes requested):
Something's not right. When I'm running with the loop interface and I open my key (whether connected to a wire or not), the circuit opens, but if I then send with the key the first character is not heard on my sounder.

AESilky:
@leskerr I will look into this shortly. I'd like to get it such that I can test on wire 11 tonight.

AESilky:
I can replicate the issue.

AESilky:
@leskerr This fixed the issue in my testing. I will try it tonight, but since I don't yet send from the key it won't be a thorough test. It will test the local sender in the station list - as, when call on, I will open with my key. I should find myself at the bottom of the station list. I can try to send 'HI' from my key as the opening. I know that's not the correct protocol, but that's all I'm comfortable in public from my key at this point.

AESilky:
On Wire 11 tonight, the station/sender list worked correctly. However, I did have the keyboard stop working. Once was my fault... I was talking about my programming experience and typed 'C++'. Of course, the '+' closed the connection. It took me a while before I figured that out. But in the other two cases, I don't think I did anything to cause it. It just stopped sending. I could still send from the key (a few random letters), so things other than the keyboard were working.

I'll look into the keyboard, but both times that it stopped working (beyond the time I typed the '+') it had worked for quite a while.

leskerr (changes requested):
I've run into two more problems with this branch:

1. When I open my key, I get a 'clack' from the speakers, but then when I send on the key, there's no audio from the speakers. And when I close the key, there's no 'click' from the speakers.

1. When I open the circuit from the keyboard (by pressing Escape, by clicking on the Circuit Closer checkbox, or by typing ~) I get a brief clack/click from the speakers and the sounder briefly drops and closes again. Then the first dot or dash from the next letter I send is missing its leading edge. In other words, typing an 'L' produces a pause followed by a 'clack', instead of 'click'-pause-'clack'.

This is all with the loop interface.

AESilky:
Good points to track down. I will look at them.

AESilky:
I have a couple questions from your document about internal operation.
> Now let's review the basic principles of operation:
>
> * Anything heard on the sounder needs to be processed by the code reader...and nothing else.
> * The key can send only if the keyboard is latched, and vice versa.
> * The key and keyboard can send to the sounder only if the internet is latched. However, they must send to the internet even if the internet is unlatched. This is so a remote station can know the state of the local system, even if the remote station's key is open.
> * When the user disconnects from the internet or changes wire numbers, the internet needs to be latched if it isn't already. If this is a new latching, then the sounder and code reader need to be notified.

Is the second bullet really the case? If I open my key-closer shouldn't that allow me to send from the keyboard? If that shouldn't be the case, then I suppose I would need to close my key-closer and open the keyboard with either ~/esc/closer-button in order to send from the keyboard. Am I understanding that correctly? If so, is that how we want it?

For the third bullet, if not currently connected then `internet_latched` is always true - right?

leskerr:
Uh-oh, @AESilky, the description you're referring to is obsolete. It was my ill-fated attempt to define an operational model that would be simple to describe and easy to implement. It went along with a roughly drawn block diagram that depicted the same thing.

My proposal fell apart for two reasons: (1) my 'latching' terminology proved to be confusing, and (2) I hadn't anticipated that an operator might want to start sending on the key and then switch mid-transmission to the keyboard.

In a discussion with @pwdirks, we decided to abandon the 'latching' jargon and go with state variables like `local_loop_active` and `internet_active` instead. @pwdirks created a new, more polished block diagram representing the revised architecture that we came up with.

AESilky:
My observations doing some testing with 2.5:

Legend:
- KEY = Physical key
- KB = Keyboard (the keyboard sender window)
- KC = Key-Closer (the closer on the physical key)
- CC = Circuit-Closer (the closer checkbox, controlled by '~'/'ESC'/mouse-click)
- PS = Physical Sounder
- SS = Simulated (Computer) Sounder
- CS = Code Sender (checkbox)
- CLOSED = The KC is in the closed position
- CHECKED = The CC or CS box is checked.
- OPEN = KC is in the open position (or no KC exists on the key)
- UNCHECKED = The CC or CS box is not checked
- CLICK = Physical Sounder is 'Down' (energized). Computer Sounder played the 'CLICK' sound.
- CLACK = Physical Sounder is 'Up' (not energized). Computer Sounder played a 'CLACK' sound.

## Loop Setting (KOB/Loop Interface):
Starting program with Loop Interface already configured, a KOB attached, and the KC CLOSED.
1. CC is UNCHECKED, CS is CHECKED
    a. Keyboard can send (of course, KEY is inactive since the KC is closed)
        1. PS & CS sound
2. Close CC (using checkbox or ESC - KC is CLOSED)
    a. CLICK
    b. KEY cannot send (loop interface and KC CLOSED electrically prevents KEY from sending)
    c. Keyboard cannot send
3. Open KC
    a. CLACK
    b. KEY can send - producing CLICK/CLACK on both PS and SS
    c. CC is still CHECKED
    d. Click CC (or press ESC)
        1. CC is UNCHECKED
        2. KEY does not sound (KS or SS)
        3. Keyboard does not sound on KS or SS (KS makes sense since KC is open and it's a physical loop circuit, SS could sound)
In 'Loop' mode, it doesn't seem to matter whether I start with the KC open or closed. The functionality seems to follow the description above.

## Normal Setting (Key&Sounder mode):
### Starting program with Normal (key & sounder) interface already configured.
#### KC Exists and is CLOSED:
1. CC CHECKED, CS CHECKED, PS CLOSED, CLICK
    b. KEY doesn't send (to be expected)
    c. KB doesn't send
2. Open KC:
    a. KC OPEN
        1. No so1und
        2. KEY not functional
        3. CC is CHECKED
        4. CS CHECKED
        5. KB not functional
    b. CC must be UNCHECKED for KEY to work
    c. With KC OPEN and CC UNCHECKED
        1. KEY works
        2. KB works
    d. Closing KC (CLOSED) the CC remained UNCHECKED, yet
        1. KEY doesn't work/send (to be expected given physical connections)
        2. KB doesn't work/send

I have more combinations to test, but my question is...
How do we want it to work.?
As a 'newbe' I have a thought, but I really want to understand how people use the program. At the same time, I would like to factor in how newbes like me might want/expect things to work.

AESilky:
@leskerr 
Rather than simply replicating 2.5 (which isn't a bad thing, given the number of years it has been around, and its user-base) - I would like to make things work logically.

There are two modes:
1. Loop (KOB) mode
2. Key&Sounder mode

There are slight differences in how to two behave.

# Commonalities
1. The KC must be open or not exist in order for the KEY to send

# Loop Interface / KOB (Key & Sounder in series)
1. in order for PS to sound, the KC must be closed (this is required to complete the electrical loop circuit through the key and the sounder)
2. The loop circuit must be energized when idle (the keyboard and internet are 'latched')
3. KEY input will directly drive PS (since it is in the loop)
4. Electronic 'Loop Interface' will allow the software to provide mute of physical or virtual output.

AESilky:
@leskerr I remember the discussions with @pwdirks and I and approved of the terminology. However, I want to assure that the  functionality is what we (new and current users) expect.

I realize that the current branch does not work correctly, but I want to make sure I understand what 'correctly' is.

AESilky:
I remember (was involved with) the discussions with @pwdirks but I'm not able to find the updated diagrams and descriptions that @pwdirks developed.

In our testing of this current change, I realize that it is not working as expected - I'm just trying to make sure I understand what is 'expected'. Now that you mention it, I believe that @pwdirks might have laid out 'what is expected'. 

@pwdirks, the previous detailed discussions, can you point me to what is expected when using either the Loop or the Key+Sounder interface from;
1. The sounder
2. The virtual (computer) sounder
3. The key
4. The keyboard
5. Physical key-closer
6. GUI Circuit-Closer ('~'/'ESC'/Checkbox)
Given the config options of:
* SOUNDER: ON/OFF
* AUDIO: ON/OFF
* INTERFACE: LOOP/KEY+SOUNDER/KEYER
(I think those are the ones that will make a difference)

Also, when starting the app with the physical key-closer open or closed in either of the two LOOP/Key-Sounder modes.

-Ed

AESilky:
I'll jot down my thoughts that I think would help a novice use MKOB
(learning Morse), yet not get in the way of a pro telegrapher that
typically uses the key.

leskerr:
@AESilky I appreciate the effort you've put into understanding how MorseKOB 2.5 works. However, when it comes to the Circuit Closer checkbox and related issues involving circuit activity, I'd rather take a fresh look and not try too hard to duplicate 2.5's behavior. KOB 2.5 evolved incrementally over the years, and I've long been unhappy with the way that aspect of the program ended up.

It would be nice to come up with an architecture whose behavior is (1) simple to describe, (2) easy to implement, (3) intuitive for new users, and (4) familiar to 2.5 users where this doesn't compromise the aforementioned goals. I'm not sure exactly how we're going to get there. It'll require some clear thinking, and I'm not sure I'm up to the challenge.

AESilky:
That is my goal.

AESilky:
To enable using a physical loop interface (KOB) I think we need two state variables in order to make things work correctly/logically.

That is to allow a user to use the GUI to 'OPEN' the circuit while the physical key-closer (KC) is closed, and to 'CLOSE' the circuit when the KC is open. As a user, these actions seem reasonable, so I think they should be supported. The reason I think we need the two state variables (rather than just the `local_loop_active` is that:
1. When the KC is open, the Sounder needs to be powered so it can follow the key. In this state, the program can't drive the PS (but can still click/clack the SS), but the KC needs to be powered in order for the KEY to drive it.
2. When the KC is closed, but the CC is open, `local_loop_active` should be true, but the PS shouldn't be driven. This allows the code to send CLICK/CLACK (drive/don't-drive) the sounder.

The difference between 1 & 2 is what makes me think we need a second state variable that tracks the KC state.

leskerr:
It doesn't surprise me too much that you'd need another state variable. My original implementation used three state variables: `kob_latched`, `keyboard_latched`, and `internet_latched`. (See the obolete description at [MKOB telegraph loop](https://github.com/MorseKOB/PyKOB/wiki/MKOB-telegraph-loop).) This was also when I thought I could have direct access to the Circuit Closer checkbox, which I later learned is only possible from the GUI thread.

When we changed the architecture and went with the `local_loop_active` temrinology, it looked like maybe that was enough, but probably I was wrong.

By the way, it still seems weird to me that the local loop could be closed when the key is open. We'll see if other users have the same reaction.

jchausler:
Hi Les and Ed,I've been trying to get my head around this as well.  I think there are may still be some inconsistencies in 2.5 but here's how I see it.  In Normal mode it seems like the circuit closer check box is in parallel with the key circuit closer whereas in Loop mode it seems like it is in series with the key (loop) circuit closer.  Of course in Normal mode, the real key and real sounder are on separate circuits, kind of parallel to each other, whereas in loop mode they are in series so that, sort of, follows.  Thus in Normal mode to send with the real key it is necessary to have the circuit closer check box unchecked otherwise being in parallel with the key it would be providing circuit continuity even with the real key open.  However, to send with the real key in Loop mode it is necessary to have the circuit closer check box checked is it is in series with the loop and if unchecked the circuit would be open regardless of the state of the key.It seems in either mode it is necessary to uncheck the circuit closer check box to send with the send window.  A difference is that in Normal mode you can send from the sender window with the real key circuit closer open (in fact as it is in parallel this is necessary as otherwise the key's circuit closer keeps the circuit closed regardless of the state of the check box) whereas in loop mode of course, the key's circuit closer needs to be closed to complete the loop.I'm not sure if this helps any but those are my observations.  Now, my question is, if the circuit closer arrangement between the real and simulated (check box) circuit closer is as above, are there still some inconsistencies in 2.5.  I'm not sure...  And of course if such behavior is what is to be provided in 4.0 does it create any inconsistencies...73, Chris-----Original Message-----

leskerr:
I hesitate to mention it, but there's a document I wrote 15 years ago that describes how MorseKOB 2.5 works. The document is [Internal Structure of the MorseKOB Program](https://drive.google.com/file/d/1HQ1Pjg_tdFQGQA6yEZl6oQslfqWbq5D-/view?usp=sharing), and the relevant section, _Operating modes_, begins at the bottom of page 3.

My head hurts every time I think about this stuff. :-(

AESilky:
@jchausler, thank you for laying out the different scenarios. @leskerr, thank you for the document link. I remember reading that some time ago, but had forgotten about it.

My goal is to make MKOB4 natural to use by an experienced user (telegrapher or ham operator) and 'easy' for a new/novice.
My thoughts...
1. Experienced User: 
 a. Does everything from the key - open/send/close
 b. Exception is key without a physical closer - uses ESC/Check-Box to open/close. Sends from the key
2. New/Novice user:
 a. Might know a few characters on the key
 b. Might open the circuit with the physical closer and send a few characters
 c. Switches to the keyboard to send more text

The combination is why I think the program needs to be able to open/close from either the physical closer or the virtual (GUI) closer. But they need to be coordinated with the current mode.

leskerr:
@AESilky I can't argue with this, other than to say experienced users often use the keyboard for one reason or another. Perhaps they don't happen to have a key attached, or maybe they want to send some text from a file.

But I thought the master branch of MKOB actually works the way you're suggesting, at least in loop mode. I'll have to go back and test it.

AESilky:
It might have. I probably broke things while adjusting to posting messages to fix other problems - so now I need to fix it.
I'm not sure that it handled the KC open and the CC controlling, but maybe it did.

I'll run a current 'master' and see how things work with it. The thing I'm trying to fix is the problems with the threading that causes things to occasionally not work.

I'll post scenarios from the current master.

AESilky:
I started doing testing with master, but then decided, that isn't the real point...
I had to make changes to use messaging to make all of the GUI work correctly. In doing so, I (most likely) broke the Physical Key/Simulated Key functionality. Rather than replicate either 2.5 or the current master 4 - I would like to implement something that experienced users and newbees will find natural.

leskerr:
Sounds good to me. Hang in there!

PR #210, opened 2020-12-27
Changed Clock.py's announcement to come out as code is generated, …

pwdirks:
…instead of completely before it

### DIFF
--- a/Clock.py
+++ b/Clock.py
@@ -119,12 +119,14 @@ def announcement(hours, minutes):
 #
 def announce(s, kob, sender, recorder, source=kob.CodeSource.local):
     global local_text
-    if local_text: print('>', s)
+    if local_text: print('> ', end='', flush=True)
     for c in s:
+        if local_text: print(c, end='', flush=True)
         code = sender.encode(c)
         kob.sounder(code)
         if recorder:
             recorder.record(code, source)
+    if local_text: print('')
 
 try:
     #log.log("Starting Clock")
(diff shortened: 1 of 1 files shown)


leskerr:
If it were up to me, I'd reverse these two lines:
```
        if local_text: print(c, end='', flush=True)
        code = sender.encode(c)
```
so each letter would be displayed right after it's sounded, instead of before. That way, you hear it first, and then get the visual display of what you just heard. Also, the timing is more consistent that way when there's a space before the letter.

But it works fine for me as is, so feel free to merge whenever you like.

AESilky:
I agree with @leskerr. 

For me to learn, I have a recording of me saying letters/numbers/punctuation and then the sounder playing the letter/number/punctuation. I also have one of 'The quick brown fox' where I say the word and then the sounder plays. I have these on a recording on my iPod that I play in my car. 

The other day I picked up my son (24) from the airport and it was playing. By the time we got home (an hour drive), he thought he knew most of the letters and numbers, but said that 'the quick brown fox was very confusing'. Unfortunately, he had learned the letters/numbers 'one-off'. He had learned 'di dah' as 'B' and 'dah di di dit' as 'C', because he heard the code and then me saying the letter. However, my recording was me saying the letter and then playing the code (explains him being one off).

Because of that experience, I plan to re-record my audio recordings such that the code is played first and then I say the letter/word.

pwdirks:
Thanks for your comments - I agree showing the text AFTER the code has been sounded would be even better.  I've changed the lines involved to read

    for c in s:
        code = sender.encode(c)
        kob.sounder(code)
        if local_text: print(c, end='', flush=True)

which I've verified plays the encoded character and then shows it on the console.

I'll go ahead and merge the change that way.

PR #216, opened 2020-12-30
Add text value to 'record' and record the decoded text

### LINKED ISSUE #214: Add decoded character to recorded information
Although it's not needed for playback, including the decoded character with the other information recorded for a code sequence would make it easer to locate a portion of a session when viewing/searching the file contents.

AESilky:
In the Reader's callback method (called once the code sequence(s) have been decoded) record the decoded text. Record it with an empty code sequence since that is already handled by the player by ignoring it.

Closes #214

### DIFF
--- a/kobmain.py
+++ b/kobmain.py
@@ -194,6 +194,7 @@ def update_sender(id):
 
 def readerCallback(char, spacing):
     """display characters returned from the decoder"""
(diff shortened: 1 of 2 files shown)


leskerr (approved):
This works great! So much more convenient for analyzing the code sequences. Thank you, @AESilky

PR #217, opened 2021-01-03
Add KOB Server URL config option to allow using an alternate server.

### LINKED ISSUE #31: Add KOBServer URL option to use an alternate KOBServer
Currently `pykob.internet.py` is hardcoded to use 'mtc-kob.dyndns.org:7890'.

Provide an option to set the KOBServer URL so an alternate KOBServer can be used (primarily for testing).

AESilky:
Added `server_url` to pykob.config. Added '-U url'/'--url url' to Configure.py. A value of 'NONE' or 'DEFAULT' results in the config.server_url value to be `None`, otherwise it will contain the value specified.

The pykob.internet class reads the URL value and if it's `None` it uses the default values (host = "mtc-kob.dyndns.org" and port = 7890). Otherwise it attempts to split the value on ':' to get a host and port value. If the URL doesn't include a port then the default port is used (7890).

Closes #31

### DIFF
--- a/Configure.py
+++ b/Configure.py
@@ -64,6 +64,7 @@ def main(argv):
     local = None
     remote = None
     interface_type = None
+    server_url = None
     sound = None
     sounder = None
     spacing = None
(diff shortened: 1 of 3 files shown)


leskerr (approved):
This works great, and it's going to be very handy while testing our new KOB servers.

PR #221, opened 2021-01-05
Update call to pykob.morse to use new parameters.

### LINKED ISSUE #220: Feed.py: gets an error 'pykob.morse' has no attribute 'AMERICAN'
The constants were changed in pykob.morse and Feed.py wasn't updated. This keeps Feed.py from running with the latest PyKOB library.

AESilky:
Closes #220

### DIFF
--- a/Feed.py
+++ b/Feed.py
@@ -57,6 +57,9 @@ Examples:
 
 Change history:
 
+Feed 1.8  2021-01-04
+- updated to use new PyKOB library
+
 Feed 1.7  2020-05-28
 - changed header to `#!/usr/bin/env python3`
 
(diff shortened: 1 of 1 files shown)


leskerr:
This works fine. I downloaded Feed 1.8 from GitHub to the new AC server, used 'configure' to set the host name, and launched a couple of test feeds to the ES server. They're running there now (see [mtc-es.dyndns.org](http://mtc-es.dyndns.org/)).

@AESilky It would be okay with me if you ever want to create a new version of Feed.py (Feed 2.0?) that uses different command line options, even if that would make it incompatible with older versions of Feed.

PR #222, opened 2021-01-05
Add ability for Feed.py to feed from a PyKOB recording (json) file.

### LINKED ISSUE #219: Feed.py: Add ability to use a PyKOB recorder file as a source
PyKOB added a recorder file in JSON format. The current `Feed.py` app can source an RSS file/feed. It would be useful for it to also accept a PyKOB recording (.json) file to use as the content to feed.

If a PyKOB recording file is specified, the content should be repeated with a short delay.

AESilky:
If path/url is a file and ends with '.json' process it as a PyKOB recording file. Otherwise process it as an RSS feed/file.

Closes #219

### DIFF
--- a/Feed.py
+++ b/Feed.py
@@ -36,7 +36,7 @@ specified number of seconds before sending again.
 Command line parameters (required):
     wire - KOB wire no.
     idText - office call, etc.
-    url - RSS formatted text source (URL) or a PyKOB recording file.
+    URI - RSS formatted text source (URI) or a PyKOB recording file.
     wpm - overall code speed (WPM).
     
 Additional command line parameters (optional):
@@ -49,7 +49,7 @@ Additional command line parameters (optional):
             (default: ignore other senders)
 
 Note: artPause and grpPause can be decimal numbers.  wire, wpm, cwpm, and
-days must be integers.  idText and url should be enclosed in quotes.
+days must be integers.  idText and URI should be enclosed in quotes.
 
 Examples:
     python Feed.py 105 "Today's News, 20 wpm, AC" "http://rss.cnn.com/rss/cnn_topstories.rss" 20
@@ -76,6 +76,7 @@ Feed 1.4  2018-07-10
 - converted from legacy morsekob module to use pykob module
 """
 
+import os
 import sys
 import time, datetime
 import threading
@@ -92,7 +93,7 @@ log.log('Starting Feed {0}'.format(VERSION))
 
 wire = int(sys.argv[1])
 idText = sys.argv[2]
-url = sys.argv[3]
+uri = sys.argv[3]
 wpm = int(sys.argv[4])
 n = len(sys.argv)
 cwpm = int(sys.argv[5]) if n > 5 else 0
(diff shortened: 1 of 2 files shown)


AESilky:
I have this running on my server against my server. Wire 125 (for Instructograph Tape 5). It repeats.

leskerr:
`Feed.py` fails for me when I specify a URL with a 'scheme' of `http:` or `file:`. For example:

```
C:\Users\Les\Documents\GitHub\PyKOB>py Feed.py 197 "Test feed, AC" "http://kob.sdf.org/morsekob/rss/beaufort.xml" 30
2021-01-06 07:33:18:    Starting Feed 1.9
  File "Feed.py", line 249, in <module>
    fileExists = filepath.exists()
  File "C:\Users\Les\AppData\Local\Programs\Python\Python37\lib\pathlib.py", line 1329, in exists
    self.stat()
  File "C:\Users\Les\AppData\Local\Programs\Python\Python37\lib\pathlib.py", line 1151, in stat
    return self._accessor.stat(self)
OSError: [WinError 123] The filename, directory name, or volume label syntax is incorrect: 'http:\\kob.sdf.org\\morsekob\\rss\\beaufort.xml'
```

It works if I just specify a path name for a local file.

A minor comment: the 'wpm' parameter has no meaning when playing a recorded JSON file, but it's required anyway. Maybe this should be optional, defaulting to something like 20wpm.

AESilky:
That's interesting. I tested with an 'http://' URL and it worked. I just tested again and it's still working. I'll use your stack trace to try to fix it, but I may need your testing. Not sure why it's different. I'll try it on my Windows machine and my server and see if I get the problem.

Yes, I thought the same thing as I was coding it. I would like to have it use the newer argument processing so we can also include some of the options. Since the wpm is currently the last of the required arguments, I can easily make it optional and default to 20. If it were an earlier one I wouldn't touch it.

Also, I tried 'breaking in' to send by opening my key and the feed kept going. It did that for both this version and the feeds you are sending to my server. I didn't try it with a 2.5 client and I didn't try it against the main server.

leskerr:
> I tried 'breaking in' to send by opening my key and the feed kept going. It did that for both this version and the feeds you are sending to my server. I didn't try it with a 2.5 client and I didn't try it against the main server.

Did you have the 'wait' parameter set to a nonzero value, to make the feed 'breakable'?

AESilky:
Ah... I forgot to do that. I'll give it a try. I don't really think I need to make tape 5 'breakable', but I wanted to test my code - it should pause the recording and then resume it when you stop. If I have a feed with tape 1 or 2 it might make sense to have it be breakable.

AESilky:
@leskerr , from PR #222 you mentioned issues with 'http://' and 'file://' URL's. For now, we won't support a 'http(s)://' reference to a recording file (should in the future), however, specifying a local file with a 'file://'URL should work.

I'm looking into that.

AESilky:
The `recorder` code doesn't support a 'file://' URI as a path (maybe it should), so I changed the code to use `os.path.isfile`. That works correctly with Windows. A 'file://' URI is supported for an RSS file.

leskerr (approved):
That works. I'll try moving my feeds to the AC server on AWS tomorrow.

AESilky:
I'll pull master and point my Instructograph feed a http://mtc-kob.dyndns.org/.

Issue #223, opened 2021-01-07
Need config option for “Automatically connect at startup”. Drop "Local Copy".


pwdirks:
The current 'config' module offers no configuration option to “Automatically connect at startup” or a “Hardware Interface” (None/GPIO/Serial port) option.  We should add them, and make sure the config/preferences dialog reads/writes these options.

pwdirks:
I guess Configure.py and the various default argument parsing should be updated, too.  Maybe with “--auto-connect=ON/OFF”?  We already have “-a” (for “audio”) and “-A” (for sounder) and “-c” (for “char. speed”); maybe “-C” for “Connect”?  There’s no “-h” or “-H” defined (we’re rapidly approaching the single-letter command option density of “ls” :-)) so I’d propose “--hardware-interface” / “-h” for the “GPIO”/“SERIAL”/“None” argument?

AESilky:
For some of the less frequently used options, I would like to see if there is a way to only support the '--' style.
Using the current arg parsing (which is very nice) '-h' and '--help' bring up the usage help. So that would rule out using '-h' for the hardware interface.

If that's possible, I think there are a few that I would move to that style and free up the '-' option for local application use while still allowing the '--' to set configuration when running the app.

AESilky:
For now, I will add both to the 'general' section of the configuration. However, 'Automatically connect at startup' is really a MKOB (app specific) configuration. It really should be in an application specific section in the config file.

However, the current `config.py` doesn't support application specific configuration settings.

Here is a snip of a current configuration file:
```
[PYKOB]
code_type=AMERICAN
interface_type=KEY_SOUNDER
key_input_invert=OFF
local=ON
char_speed_min=18
remote=ON
server_url=kobserver1.aesilky.net
sound=OFF
sounder=ON
spacing=NONE
station=Instructograph Tape 5.2
wire=125
text_speed=28
```

As can be seen, all of the config values are in the 'PYKOB' section.
There could be (should be) a 'MKOB' section where the 'Automatically connect at startup' value would be stored.

That can be treated as a separate issue, because other config storage changes might make it irrelevant.

AESilky:
Until/unless we can figure out how to add '--' only options, the 'Hardware Interface' probably needs to be '-H'. As mentioned, using the standard option parsing uses '-h' to bring up command help.

AESilky:
[...]

Issue #117, opened 2020-07-16
Trap invalid characters in numeric fields


leskerr:
MKOB produces a system error if the user accidentally enters a nonnumeric character in the WPM or Wire No. fields in the GUI window. Although the program keeps running, this situation should be handled more gracefully.

AESilky:
I also ran into a 'divide by zero' error when I was changing the speed from 30 to 20. With the value of '30' currently in the input I clicked between the '3' and '0' and hit 'backspace' to remove it. My goal was to then enter a '2' to change from '30' to '20' - but when I hit backspace (removing the '3') I got a DBZ error and the app stopped working.

It's typical for applications to ignore input until the focus leaves the input field. This would resolve this problem. If the focus does leave the field with a '0' value, then the value can be checked at that point and a message displayed.

leskerr:
I agree. The input should be ignored until focus leaves the input field. Good suggestion!

Of course, if the user clicks on one of the up or down spinbox arrowheads, that needs to be processed immediately. But that's a different thing.

AESilky:
Yes. In that case the value should be valid so it can be applied immediately. The wire should have a check to not let it go below 1 (is there an upper limit?). The speed should have a lower and upper limit (maybe it already does).

leskerr:
The WPM widget is currently limited to values ranging from 5 to 40. Wire numbers go from 1 to 32000. These limits constrain the up/down spinner arrows. I'm not sure what happens if you type in a value outside that range. Maybe we need a separate check in that case.

AESilky:
This was fixed some time ago. Not sure of the exact version.

PR #226, opened 2021-01-12
Add new pykob/preferencesWindow.py and incorporate GUI prefs/config options in Configure.py

pwdirks:
Here's a first take on adding a config dialog.  I'm open to changes to the file name ("preferencesWindow.py"), especially because I gather on Windows what Mac people call "Preferences" is often called "Options"?

The guts of the prefs dialog are in the new pykob/preferencesWindow.py but I've also added a "-G"/"--gui" option in Configure.py so you can now run Configure.py just bring up the same preferences dialog.

Finally there's some trivial cleanup in a few places, cleaning up trailing spaces on lines etc.

IIRC we still need to settle on proper terminology for "dot length" (="Code speed" in current dialog) and how quickly characters come after inserting Farnsworth spacing ("character rate" in this version).  FWIW I chose "code speed" because it made sense to me as the speed of dots and dashes in a character.  "Learning to read code at 25 WPM" certainly suggests a certain speed (=dot length).  I figured "character rate" was good for the rate at which characters came at you (one/sec for the equivalent of 12 WPM etc.) but all suggestions welcome!


leskerr:
@pwdirks When I click on the Save button on the Preferences panel, the program terminates. That's not how it's supposed to work, is it?

By the way, I don't care whether it's called Preferences, Options, or Settings. It's all the same to me.

pwdirks:
> On Jan 11, 2021, at 11:44 PM, Les Kerr <notifications@github.com> wrote:
> 
> 
> @pwdirks <https://github.com/pwdirks> When I click on the Save button on the Preferences panel, the program terminates. That's not how it's supposed to work, is it?
> 
> 
Haha - no!  That’s a new behavior due to a change that worked well in the Configure.py case (where it WASN’T quitting the main loop, and leaving an empty GUI running) but I didn’t re-test in MKOB.  Let me update the branch to revert to the old behavior while I sort that out.

73, Patrick.

leskerr:
It's working now, and this is really fun! Here are a couple of initial reactions:

- Changes made on the Preferences panel don't seem to take effect until you relaunch the program. This isn't what I expected. If I want to turn the sound off, for example, I'd want that to take effect immediately.
- I'd like the text box for the station ID to be a bit wider...maybe the same width as the server text box.

pwdirks:
> On Jan 12, 2021, at 2:23 PM, Les Kerr <notifications@github.com> wrote:
> 
> 
> It's working now, and this is really fun! Here are a couple of initial reactions:
> 
> Changes made on the Preferences panel don't seem to take effect until you relaunch the program. This isn't what I expected. If I want to turn the sound off, for example, I'd want that to take effect immediately.
That brings up an interesting question when users have changed something in the main panel, too.  If the running state is different from what was configured for use at the start, what would you expect to be the setting when you dismiss the dialog box?  ... and does that expectation change depending on whether you actually modified it in the preferences panel or not?

To take your example, if your configuration specifies “Code sender” OFF but you turn it on in the main window (so the “running state” = ON) and you bring up the Preferences dialog and hit “Save” - would you expect the setting of “Code sender” to suddenly reflect what’s configured for startup?  Or only if that’s different from the current running state?  Or [worst of all, from an implementation perspective] only if you’ve touched that setting in the Preferences panel and changed the config?

Kidding about implementation difficulty (due to the separation from the main program and the innards of what the user clicks on in the Preferences panel) aside, I think making it depend on whether or not you touch it is probably a bad idea.  What if you tweak it but set it back to the previous value?  What if you leave it alone because it looks set correctly?
> I'd like the text box for the station ID to be a bit wider...maybe the same width as the server text box.
I’ll take care of that in the next iteration.

Thanks!
-Patrick.

leskerr:
Patrick,

I was expecting changes to the preferences panel to be immediate and permanent, while changes to the main panel would be temporary. For example, if I change my station ID on the preferences panel and click on the Save button, the change would take effect immediately (other users on the wire would see my new ID), the new station ID would appear on the main panel, and the new ID would be used the next time I launched the program.

On the other hand, if I change the ID on the main panel, it should take effect immediately but on relaunching the program the old value would persist.

In other words, all changes made on the preferences panel are immediate and permanent (assuming you Save and don't Cancel), while changes on the main panel are immediate and temporary. I was hoping to try this functionality for a while and see how it works in practice.

~Les

AESilky:
I'll pull this branch and give it a try.

pwdirks:
Hi Les,

So, when you bring up the Preferences panel, which is initialized from the values in the config file, not the running values in the main window, and you hit “Save” you expect all those settings to take immediate effect?  That, essentially, the main window resets to the state it would have on the next startup, with all configuration values as set in the config file?  And if you opt for “auto connect at startup”, should that latch the internet when you exit?  That would apply to all the main window-controllable settings?

  - Station ID
  - [initial] Wire #
  - Speed (WPM)

Or, of the ones that have a config, maybe to some but not to others?

73, -Patrick.

leskerr:
It's a dilemma, isn't it? It just seems very awkward that you have to close the program and start it up again whenever you want to change a setting, especially if you're in the middle of a session and you want to do it on the fly.

pwdirks:
Hi Les,

I don’t see a problem with re-reading the config after launching the config panel - that’ll have the effect of having all changes take immediate effect.  That’ll make an easy way to, for instance, change the serial port and start working with the new port.  I agree having to quit and restart should never be necessary.

I’m only concerned about the areas where there might be overlap between the configuration in Preferences and the (temporary) changes made in the main screen (as is the case for station ID, wire number, and speed).

I suppose of those we could arbitrarily reset some but not others (e.g. change the station ID to the configured one but leave WPM unchanged from the main screen)?  That’s starting to sound a bit weird, too, though.

OTOH, any changes of those three (station ID, wire #, and WPM - am I missing anything?) we decide to make “sticky” (=config is immediately changed to reflect them) would also mean the preferences panel starts up with the identical values shown in the main screen and then I think it wouldn’t be strange to copy them out of the config after you hit “Save”.  If you change, say, WPM in the config dialog it’ll immediately show that new value in the main screen.

Much as I prefer the idea of making some config settings (wire #, code speed), configurable as “INITIAL settings” and leave the main window transient, it may be easier to model it as if there’s only ever one setting in the system for wire #, WPM, Station ID, etc.

73, Patrick.

leskerr:
Patrick,

I'm starting to come around to my original way of thinking, that none of the settings on the main screen should be duplicated on the preferences panel, and any changes to the main screen should be permanent. (Do I hear cheers of agreement from @jchausler and @wamorganjr?)

If you can implement this, I'm anxious to give it a try.

~Les

wamorganjr:
I will second the motion. 

Chip N3IW 


-------- Original Message --------

pwdirks:
Hi Les,

Well, if 100% of the main screen settings are sticky (that is, immediately reflected in the config file for the next run), then I think it won’t look strange to have ALL config options available to set in the Preferences panel, since they’ll start from the exact same values as the current main screen.  If you change the station ID while you’re using MKOB, that’s the station ID that’ll show up in Preferences.  Given that, if you change it, I think the same consistency should be applied in reverse and the change should immediately be reflected in the main screen.

I’m concerned forcing users to remember what can be set (only) in the main screen and what can be set in the Preferences panel will lead to frustration.  Want to change the serial port?  Go to Preferences.  Want to change from “Loop” to “Key + Sounder”?  Go to Preferences.  Want to change to turn off local sound?  Nope, that’s only in the main window?

Let’s try it where there’s never two different ideas about a given setting.  That means

  - making all main window changes immediately “sticky” (you’re welcome, Chris :-))
  - making all configurable settings available in Preferences
  - making any changes in Preferences immediately effective in the main window

That also avoids having to craft a special “Preferences-Plus” window for use in “python3 Configure.py --gui_mode” (where there IS no main screen to confuse things, and everything should be settable).

I’ll push an update on the branch when I have that behavior available for testing.

73, -Patrick.

leskerr:
Patrick,

Your latest proposal works for me.

> Let’s try it where there’s never two different ideas about a given setting. That means
> - making all main window changes immediately “sticky” (you’re welcome, Chris :-))
> - making all configurable settings available in Preferences
> - making any changes in Preferences immediately effective in the main window

Don't forget to include the Circuit Closer in the settings to be saved from the main window. And looking down the road, there are the window size and window position values as well.

~Les

jchausler:
Yes, once again all configuration options should be only set in one place and the WPM, wire number and station ID should be set on the main GUI page as they are in 2.5.  I can see no valid reason for another approach.  By having them set in multiple locations and each location having a somewhat different impact is in my mind way over complicating the issue.  In designing software I've always believed in the KISS principle and user interfaces need to be simple and clear and provide obvious behavior.  This is particularly true for what I expect is the primary audience for this program.  If we really desire special behaviors, possibly special versions of the program can be provided.  I recall when Les first set out on this journey he referred to creating a telegraphers toolkit for building such applications.

73, Chris

pwdirks:
Well, as Einstein I believe said, “Everything should be made as simple as possible, but no simpler.”

Let me see if I can’t strike a reasonable balance.  I’ll let you know when I can push an update to this change.

73, -Patrick.

AESilky:
After listening to the issues, I can agree with making the main window settings 'sticky', but don't agree with Chris on the 'only one place'. I think all options should be settable in the configuration panel (just as they can be set in the command line `Configure` app). If I go into the configuration panel to set something, I may decide that I also want to set the WPM value (because I changed 'spacing' for example).

AESilky:
I'm a little hesitant on `Circuit Closer` being a preference and being saved. I'm currently in the middle of circuit closer code changes and they have to do with the current state of the physical key closer. I don't know that just because I ended with the key open I will want to start with it open.

My current changes always start the app with the virtual closer closed and then check the physical closer. If the physical closer is open it assumes the key doesn't have a closer and no longer checks for it - leaving the virtual closer closed, but now in total control.

AESilky:
There is also an issue open for doing a number check on the values in the main window before trying to use them. If I'm changing the wire value, I regularly delete the value to type in a new one (say if I'm going from 107 to 11). When I do that, an exception is thrown when I delete the value because it is now blank and the code tries to do an 'int' conversion on it.

I also don't like that as I type in the Station ID is uses each value as I type it. I end up with 'myself' registered with many (partial) station IDs until the old/incomplete ones time out.

leskerr:
> I'm a little hesitant on `Circuit Closer` being a preference and being saved.

I mentioned the Circuit Closer because MorseKOB 2.5 currently saves that setting between sessions.

jchausler:
I don't know all the differences in the on screen circuit closer on 4 as compared to 2.5 but in 2.5 if you are in loop mode, you normally want the on-screen circuit closer closed, it seems to be more or less in series with the real circuit closer.  However in sounder driver (normal) mode, you want the on screen circuit closer open as it appears to be in parallel with the real circuit closer so if it is closed and you open the real circuit closer and start sending, nothing happens as the circuit is still closed.  So to some extent the circuit closer setting also should be sticky.

As to the station ID and the wire number, I normally don't change either with the program connected.  I first disconnect and then make any changes, then reconnect.  If the wire number field is blank, internally think of it as zero so you won't get an exception.

When changing wire numbers I frequently use the up down arrow keys as well as editing the window directly but try to remember to first disconnect.

73, Chris

leskerr:
> There is also an issue open for doing a number check on the values in the main window before trying to use them.

I agree, this is an important issue. The problem I constantly run into is when I enter a wire number and later type something, thinking it's going into the keyboard area, and the program crashes because of an invalid wire number.

AESilky:
I really like the input on the closer from veteran users, however, in my use during testing with;
1. Loop with KOB that has a physical closer
2. key&sounder with a key that has a closer
3. key&sounder with a key that doesn't have a closer
I think the 'application/virtual) closer should come up closed. That allows checking the physical closer to see if it is open. That means that the key doesn't have a closer.
If it doesn't come up as you expect, that means that you probably have your (physical) closer open when starting the app.
To me, that makes things predictable and easier to trouble-shoot.

AESilky:
All things considered, I would like to implement something and try it out and see what people think.
That is the main point of agile development... Put it out there and see what people think. If they love it you are good to go. If they complain, then you change it.

AESilky:
So, I don't think I would want 'connected' to be a saved configuration value. It currently isn't a 'config' value and isn't supported in `Configure`, but maybe it should be.

AESilky:
Issue #117 is for invalid values in the Main Gui input fields.

Use case is:
I delete the value in 'wire' in order to change from '105' to '11'. On deleting the current value I get a 'value' exception and the app crashes.

AESilky:
I mention this, because it should also be handled in the preferences panel. The values in the fields should not be analyzed until 'save' is used. If there are invalid values ("My preferred rate" for 'WPM') - how is that handled?

AESilky:
@jchausler please provide a use case where we want the 'closer' value to be persisted and applied the next time the app is run.

jchausler:
Actually, I don't want the closer value to be persistent.  The default practice should be to come up disconnected.  What I think we should consider is an option somewhere on the config menu, and I would suggest at a lower level, which says the program should come up connected.  The only reason for this is a situation which occurred only once at the AWA museum when a docent started up the telegraph PC, clicked on one of the MKOB 2.5 icons on the screen (there are two as two circuits) and then wondered why things didn't start "clicking".  Once I explained to him that he had to also click on connect there was no further problem.  But expanding on this, if there is an unattended installation where all that is done to start the display is to turn on the power, we need to both have the application in some manner start automatically (I would say put it in the autoexec.bat file which shows you my age and experience, but what I did with the Mill :-) and for it on start up to connect to whatever wire was specified in the wire number selection when it was previously shut down.

pwdirks:
Hi Chris,

I agree and the current design already has a checkbox labeled “Automatically connect at startup” for exactly this reason.  If that’s checked there’s no need to click on “Connect” when the program starts.

There’s not yet a config option to reflect this but we’ll introduce one alongside all other config options to it can be set like all other options.

73, -Patrick.

leskerr:
> All things considered, I would like to implement something and try it out and see what people think.
> That is the main point of agile development... Put it out there and see what people think. If they love it you are good to go. If they complain, then you change it.

I agree with this. The thought experiments of trying to imagine how things would work are too difficult for me. I'd much rather try something that's been implemented and see how it works in practice.

AESilky:
Issue #223 adds the auto-connect option value and the functionality in MKOB. The changes just need to be merged into master.

AESilky:
@pwdirks I want to test this out, but get the merge conflicts when I pull it. Should I just fix those locally for now so I can test the preferences dialog, or are you close to doing that and pushing a new copy?

pwdirks:
The main thing missing in the Preferences window right now is having the preferences take effect in the main window immediately after you hit “Save”.

We could just file a separate issue for that, and merge the Prefs code as it stands now?

I have no changes on my local branch right now - if it’s simpler for you, feel free to merge the current ‘pull’ request and resolve the conflicts you hit?

73, Patrick.

AESilky:
Okay. Let me merge locally and try it out. If all seems to work (other than not taking effect immediately), I can push the merged version and accept it.

I'm fine creating another issue for having it take effect.

pwdirks:
The alternative is to merge it on the server, update your master, sort out the incompatibilities, then merge your resolved change?

I’ll leave it to you to pick whatever’s simplest for you.

73, -Patrick.

AESilky:
I notice that (on Mac) the window starts in the top-left and then jumps to the center. Not something to keep from moving forward, but not what I would expect. I will try it on Windows and Linux.

AESilky:
Looking at the Preferences dialog, I have some comments, but they are minor - so I think we should take it and then make changes as needed. I'll do a bit more testing and then push the merge (just a minor conflict that was easy to resolve).

That way, we can also accept my 'config' change for the 'Auto-Connect' and get both of the PR's and branches out of the way.

pwdirks:
Hi Ed,

There may be ways around that or mitigate the effect - either ‘iconify’ or ‘withdraw’ the window when it’s created (there seems to be no way to create it invisibly), add all the content, then reverse that to make it visible again.

Would you mind filing a separate issue to pursue that?

Thanks,
-Patrick.

AESilky:
I changed various settings and checked with (CLI) `Configure.py` and they all seem to be reflected. I'm going to push my (minor) merge so we can use this and create separate issues as necessary.

@pwdirks Thanks for the great work on this!

PR #229, opened 2021-01-15
Remove dependency on kobconfig; pass through all config options straight from 'config'.

pwdirks:
This change removes all references to "kc" (kobconfig.py); all config options are directly taken from config (pykob/configure.py).  Changes to wire #, station ID, and WPM were already passed through directly to config file; this code adds a guard to make sure the values of wire # and WPM are valid integers before returning them (otherwise substituting "1").

This is a first step to incorporating the Preferences dialog - this change makes sure the config file is in sync with the main panel at all times and that the PyKOB configs are the only interface to the config file(s).


leskerr:
It works for me, Patrick. I notice the main window starts up smaller than before, but that'll be OK once the window size and position are saved for the next time the program is run.

~Les

pwdirks:
Actually I noticed the window size and position were taken from fixed constants in kobconfig.py; I figured that was a shortcut to getting the actual window size so I "fixed" the code to compute the actual window size and use that directly instead.  The size and position are not currently saved in the config file anywhere (they were hard-coded in kobconfig.py) so it should've always been coming up in the same place, and at the same size.  Now it comes up in its minimal size, centered horizontally, a bit above the center of the screen (there's a 40/60 split of the vertical padding built into the code).

Let me know if you want to revert to using the previous hard-coded window size (and why).  If it's important (e.g. for some text field to be a different size) we may be better off specifying that in the text field creation and otherwise let the size be whatever it needs to be.

leskerr:
Patrick, there's nothing magic about those size and position numbers in the old, obsolete code. If you can make the window the same size as the MorseKOB 2.5 window, it might make it appear more familiar (i.e. 'comfortable') to existing users, but don't go to too much effort on that. Centering the window on the screen initially is fine. The size is a bit cramped as it is, but it can always be resized (and repositioned) on the fly by the user.

At some point we'll want to save the size and position between sessions. This will be especially important when the user creates multiple profiles, with the respective windows coming up in preestablished locations on the screen when each instance is launched.

~Les

AESilky:
This is probably why my #223 change has a conflict. I think it's fine, I'll prefer your changes over mine when I merge. The main point of my #223 was to add the auto-connect option and use it in MKOB to connect at start up. But I also came across the `kobconfig` redundancy and removed them - just leaving the size and location values.

AESilky:
The config file should have an application specific section where things like the size and position can be saved. The config file is a traditional Windows format that has [section-name] sections. There should be a [mkob] section that has values specific to the MKOB app.

AESilky:
`Configure` should only set/show the common configuration. Any app specific configuration should only be used/set by the app. This is normal Windows config file behavior. It's a bit different for Mac and Linux, but since the app started as a Windows app, that is how things work.

AESilky (approved):
I'll update my #223 to favor these changes.

jchausler:
I'll second the goal of having size and position recalled between sessions.  At the AWA I normally have two sessions of 2.5 running, one in the upper left of the screen and the other in the lower right when they are not in full screen mode.  This even allows a little overlap as I'm normally working with the one on the lower right side of the screen and I overlap its window with the one in the upper left as it is usually just a broadcast wire and unless I need to change them, hiding the controls on the lower right of the window still leaves the reader screen portion uncovered.  When I exit full screen mode they return to their former size and position.

73, Chris

pwdirks:
Les, see issue #233 to make the main window’s widgets explicitly the size in the current display (fixed at 800x500) by changing the size of various text widgets to natively make the window and its contents about the same size (802 x 503, IIRC).

I just put out a pull request for a change to do so (and in the process hide the window during construction so it doesn’t “jump” quite so visibly at startup).

The task of saving the size and position across invocations remains for a separate change.

73, -Patrick.

PR #234, opened 2021-01-16
Remove fixed 800x500 main window size; clean up use of 'root' vs. 'self.root', and minimize window visibility before final placement

pwdirks:
This change makes the window size as close as possible to the current size (802x503 vs. 800x500) by explicitly specifying the sizes of various fields rather than letting the "sticky=NSEW" determine how much to "stretch" every widget to make it fit.  

This change also uses root.withdraw() / root.deiconify() to hide the window while it's being contstructed and before its final placement.

Finally this change cleans up the mixed use of "root" (a parameter) and "self.root" (where it's saved) during use.  Either should do; this change alters it to consistently use "root." for brevity.  In addition some trailing white space was removed from a few code lines.

### DIFF
--- a/kobwindow.py
+++ b/kobwindow.py
@@ -32,7 +32,7 @@ Create the main KOB window for MKOB and lay out its widgets
 import kobmain
 import kobevents
 import tkinter as tk
-import tkinter.scrolledtext as tkst 
+import tkinter.scrolledtext as tkst
 import kobactions as ka
 import kobstationlist as ksl
 import kobreader as krdr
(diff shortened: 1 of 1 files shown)


leskerr:
Patrick, the size is larger now, but the window appears briefly in the upper left before jumping to the center of the screen. It doesn't seem too important for now, but I think this is something you were trying to fix. This is with Windows 10.

~Les

AESilky:
This seems straightforward and I wanted to merge it so I could have a clean shot at merging my config changes.

PR #238, opened 2021-01-17
Added try ... except around attempt to convert wire number with int(.…) in pykob/config.py:764

pwdirks:
Fix seems to only require adding a try ... except guard around the conversion attempt in pykob/configure.py.

### DIFF
--- a/pykob/config.py
+++ b/pykob/config.py
@@ -761,7 +761,11 @@ def read_config():
         __key = __WIRE_KEY
         _wire = user_config.get(__CONFIG_SECTION, __key)
         if (_wire) or (_wire.upper() != "NONE"):
-            wire = int(_wire)
+            try:
+                wire = int(_wire)
+            except ValueError as ex:
+                log.err("Wire number value '{}' is not a valid integer value.".format(_wire))
+                wire = 1
     except KeyError as ex:
         log.err("Key '{}' not found in configuration file.".format(ex.args[0]))
         raise


pykob/config.py
```diff
@@ -761,7 +761,11 @@ def read_config():
         __key = __WIRE_KEY
         _wire = user_config.get(__CONFIG_SECTION, __key)
         if (_wire) or (_wire.upper() != "NONE"):
-            wire = int(_wire)
+            try:
+                wire = int(_wire)
+            except ValueError as ex:
+                log.err("Wire number value '{}' is not a valid integer value.".format(_wire))
+                wire = 1
```
AESilky, line 768: Agree. Odd thing is, that I thought I had that - obviously not. Only difference is that I set the wire to 0.

AESilky (approved):
Looks good.

pwdirks:
Hi Ed,

I first defaulted it to “1” but thought about making it “0” as well.  I almost changed it but then figured, in case somewhere something decides “0” is not a valid wire number, “1” would be safer.  I don’t feel strongly - feel free to change it to default to “0” in another change if you like.

Thanks for merging this pull request.

73, -Patrick.

AESilky:
Only thing is that `wire` is initialized to `0` when the class is created.

I'm working on the config class as part of 236, so I can change it as part of that. There are a few things that need to be 'fixed' in that class to implement 'observer'.

Issue #235, opened 2021-01-16
MKOB: Preferences doesn't set 'Auto-Connect' value in Config.


AESilky:
If I set the 'Auto-Connect' value using the CLI Configure.py and start MKOB it does/doesn't auto-connect as expected. However, the value doesn't show correctly in the 'Preferences' dialog and changes in the dialog are not made to the configuration.

pwdirks:
Fixed with merge of branch #247 from MorseKOB/pwd-#235-Incorporate-new-auto-connect-config-in-preferences-panel

PR #239, opened 2021-01-17
Wire up auto-connect option in Preferences dialog.  Removed some extraneous print() output at same time.

pwdirks:
"auto_connect" option was new since "Preferences" dialog was added and the preferences dialog used a fixed value of "0" for its setting.  This change updates the preferences panel to get/set it in the config from the actual config option.


pwdirks:
Not needed; change has been merged from 'pwd-#235-Incorporate-new-auto-connect-config-in-preferences-panel' branch.

Issue #243, opened 2021-01-19
Preferences panel should re-present same panel if a Preferences panel is already up


pwdirks:
Clicking on File -> Preferences and then clicking on the main window and re-selecting File -> Preferences brings up a second preferences dialog along with the first one.

Instead the code should merely bring the first one to the front if a preferences dialog is already displayed.

pwdirks:
Fixed by merge of pull request #244 and latest change on branch pwd-243-limit-to-single-preferences-panel-at-runtime.

PR #244, opened 2021-01-19
Pwd #243: limit to single preferences panel at runtime

pwdirks:
Change preferences logic to rely on running Tkinter mainloop() instead of starting a separate one; track preferences window and merely pull it to the front with root.lift() instead of creating another one.  Add support for a callback in the preferences window when it's dismissed (with 'Cancel' or 'Save') to help track the liveness of the preferences window.

### DIFF
--- a/kobactions.py
+++ b/kobactions.py
@@ -41,6 +41,8 @@ import kobstationlist as ksl
 import pykob  # for version number
 print("PyKOB " + pykob.VERSION)
 
+global preferencesDialog
+preferencesDialog = None
 kw = None  # initialized by KOBWindow
 
 ####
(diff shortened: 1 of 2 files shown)


AESilky (commented):
I'm going to pull this and test now.

pwdirks:
I just updated the 'pwd-243-limit-to-single-preferences-panel-at-runtime’ branch to include a change to also de-iconify the window when it’s brought to the front.

What’s the easiest way to get this in, too?  If I create another merge request for the same branch, will that cause trouble or will it correctly incorporate the delta now on that branch?

Please try the new branch and let me know how you’d like to proceed?

Thanks and 73, -Patrick.

P.S. I’m not sure what the Mac equivalent of “iconify” is but I guess that’s the “minimize to the Dock” orange button in the window frame.  I checked you can hide the window using that and it appears front-most, normal size, if you hit File -> Preferences again.

AESilky:
Just push it. I left the server branch open, so that will allow a new pull
to be created.

I just tested and it works fine.

-Ed

pwdirks:
I’ve pushed the branch up to ‘origin’.  Try merging the fix?

73, -Patrick.

AESilky:
I haven't tested on my mac. I have changes open, so it's work to stash them and check there. I (kinda) figured you would test that. I can do a test on Linux.

AESilky:
This works for me. You will need to open a new PR though (from the changes) so it can be merged.

pwdirks:
Mac is the easy test for me - that’s where I do development.  It works on macOS.

pwdirks:
OK - see  #245 (If the preferences window is minimized ('iconified'), selecting 'Preferences' does not show the window as expected).

AESilky:
Yes, me too (Mac is main dev machine). Only go to Windows machine for testing and Win specific issues. Just that right now I have changes in my Mac env that I would need to 'stash' in order to test this.

PR #246, opened 2021-01-20
Limit preferences to a single window; de-iconfify and bring the preferences dialog to the front if already on-screen

pwdirks:


pwdirks:
Closed w. merge of latest change on branch pwd-243-limit-to-single-preferences-panel-at-runtime.

PR #251, opened 2021-01-21
Pwd 250 guard optional library inclusion and use

pwdirks:
Updated references to Tkinter to be under guard; changed Configure.py to also disable -G option if no GUI support is available.  Changed pykob/preferencesWindow.py to not offer serial configuration if pySerial is not installed, and stub out any attempt to create a GUI window if Tkinter is not available.

No attempt to conditionalize Tkinter use in MKOB and its related source files (kobactions.py, kobwindow.py etc.)

### DIFF
--- a/Configure.py
+++ b/Configure.py
@@ -31,7 +31,12 @@ Provides a Command Line Interface (CLI) the the pykob.config module.
 """
 import argparse
 import sys
-import tkinter as tk
+GUI = True                              # Hope for the best
+try:
+    import tkinter as tk
+except ModuleNotFoundError:
+    GUI = False
+    
 from pykob import config
 from pykob import preferencesWindow
 
@@ -102,7 +107,9 @@ def main(argv):
             config.station_override, \
             config.text_speed_override, \
             config.wire_override])
-        arg_parser.add_argument('-G', '--gui', dest="gui_config", action='store_true', help="Use preferences panel GUI for interactive configuration.")
+        if GUI:
+            arg_parser.add_argument('-G', '--gui', dest="gui_config", action='store_true',
+                            help="Use preferences panel GUI for interactive configuration.")
 
         args = arg_parser.parse_args()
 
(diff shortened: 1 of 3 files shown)


AESilky:
No need to conditionalize the use of tkinter in MKOB and it's files. MKOB is a GUI app and therefore requires tkinter. Note that MKOB and its support files used to be in a MKOB subdirectory, making it obvious that they were all part of the MKOB app. However, that made it very difficult to develop and test when changes needed to be made in the 'pykob' module files. We pulled all of MKOB up out of the subdirectory so that 'pykob' was a subdirectory. That allows changes in those module files to be picked up without having to publish the module.

The 'pykob' module files are not graphical and can be used by both MKOB and command line apps.

A convention that Les picked and I've stuck with is that application files (meant to be run) start with a capitol letter while support/module files start with a lowercase letter. In addition, the MKOB files start with 'kob' (in addition to the main MKOB.pyw).

pwdirks:
> On Jan 21, 2021, at 11:03 PM, Ed Silky <notifications@github.com> wrote:
> 
> 
> No need to conditionalize the use of tkinter in MKOB and it's files. MKOB is a GUI app and therefore requires tkinter. Note that MKOB and its support files used to be in a MKOB subdirectory, making it obvious that they were all part of the MKOB app. However, that made it very difficult to develop and test when changes needed to be made in the 'pykob' module files. We pulled all of MKOB up out of the subdirectory so that 'pykob' was a subdirectory. That allows changes in those module files to be picked up without having to publish the module.
> 
> 
Yes - I noticed that organization and I think it works well - a bunch of top-level programs in the top-level directory sharing a single “pekob” subdirectory for the common core functionality.

And, yes, I figured the same - MKOB is a GUI program; no point trying to run it w/o Tkinter installed.
> The 'pykob' module files are not graphical and can be used by both MKOB and command line apps.
> 
Right, so I concentrated on making sure anything in pykob (and, for good measure, the Configure program itself) can work without Tkinter installed.
> A convention that Les picked and I've stuck with is that application files (meant to be run) start with a capitol letter while support/module files start with a lowercase letter. In addition, the MKOB files start with 'kob' (in addition to the main MKOB.pyw).
> 
I noticed that and it works for me.  The pieces of MKOB also all seem to start with “kob...” Which is helpful as well.  I think all these conventions work well.

Issue #249, opened 2021-01-20
Config/preferences should be split into General/basic and a separate "Advanced" tab


pwdirks:
The current preferences dialog shows all available config options.  Some, like the server URL and port, are better moved to a separate "Advanced" tab to avoid unnecessarily cluttering up the basic preferences display.

pwdirks:
I propose we move:

* Invert key input (only used for modem input)
* Server address and port number 

to the advanced tab.  We could also move "Word spacing" - does anyone tweak where Farnsworth spacing is introduced, or is "between characters" the overwhelming favorite?

Should we leave "Code type: American/International" in basic preferences or move it, too, into "Advanced"?

I'm pretty comfortable calling one tab "Advanced" but what should we call the other, which is first shown?  "General"?  "Basic"?  "Common"?  Something else?

leskerr:
Patrick,

The serial port pull-down menu, interface type (loop, key & sounder, or keyer), and audio on/off all need to be on the basic tab. Everything else can move to the advanced tab. Perhaps a separate Farnsworth tab is in order.

And yes, character spacing for Farnsworth is definitely more common.

~Les

AESilky:
I agree, and would add 'sounder' on/off to 'General'.

pwdirks:
Hi guys - have a look at the pwd-#249-Create-tabbed-preferences-with-separate-Advanced-tab branch sometime.  I took a first stab at dividing preferences into "Basic" (maybe we should call it "General", or something else?) and an "Advanced" tab.  I created a separate third tab with all the "Morse Code" options.

Comments about placement of various options (and anything else, of course) would be most welcome.  It's not hard to shuffle things between tabs.

leskerr:
Patrick,

I love the tabs!!

![Preferences](https://user-images.githubusercontent.com/63867213/105617822-3f18ce80-5d96-11eb-9be4-8bd7b69c1c15.png)

Here are some comments:

1. I'd move everything in the 'Internet Connection' panel to the Advanced tab. I can't imagine using any of these options myself, and I'd rather not confuse first-time users with features they're unlikely to use.
- I don't know why I'd want to suppress sending to other stations while I'm connected to a wire. Is it because I want to monitor the wire but not accidentally send something? What's the use case?
- You'll have to tell me one more time what the intial station ID and initial wire number are for. Changes to these settings on the main screen are 'sticky', so they're also used as the 'initial' settings. Why are they in both places?
[...]

PR #252, opened 2021-01-24
Pwd #249 create tabbed preferences with separate advanced tab

pwdirks:
First complete version of a tabbed preferences window.  Items are easily shuffled between tabs - please look at which preferences are where and let me know if you'd recommend changes?.  Use of a separate "Morse Code" tab is optional, but I think it's convenient.

### DIFF
--- a/pykob/preferencesWindow.py
+++ b/pykob/preferencesWindow.py
@@ -55,6 +55,19 @@ class PreferencesWindow:
 
         self.root.title("Preferences")
 
+        
+        #######################################################################
+        #
+        #   Create two-tabbed interface: Basic/Advanced
+        #
+        #######################################################################
+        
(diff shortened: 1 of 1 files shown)


leskerr (approved):
Patrick,

The revised preferences panel looks nice and works well. Having a separate 'Morse Code' tab is exactly the right thing to do.

Great job!

~Les

AESilky (commented):
I won't have a chance to review this until late tonight or tomorrow morning. If @leskerr has had a chance to run it, I'm good. If I notice anything when I have a chance to look, we can just do a separate fix for it.

pwdirks:
Sounds good to me.  Please do have a look sometime (no rush) - moving things around between tabs is pretty trivial.

I’ll merge this sometime this afternoon.

Thanks and 73, -Patrick.

PR #262, opened 2021-02-04
Fix #260 by changing 'character rate' to 'dot speed' …

### LINKED ISSUE #260: Revise Farnsworth terminology
I find the terminology currently used on the Morse Code tab of the Preferences panel confusing. I suggest we try 'Code speed' and 'Dot speed' instead.

### LINKED ISSUE #261: MKOB: Implement 'Use local sounder' setting
Currently this checkbox on the Preferences panel has no effect.

pwdirks:
… and selectively enabling/disabling spacing options depending on selected speeds.

This change re-labels "character rate" to "dot speed" as an experiment.

In addition to the purely cosmetic change of label, the code now also check to see if the "code speed" and "dot speed" are the same and disables Farnsworth spacing selections, fixing the value to "NONE".  Conversely, as soon as the two speeds diverge, "NONE" is deselected and "Between characters" and "Between words" become enabled, with whichever one was originally configured selected.

Try this out and let me know what you think of the effect?

### DIFF
--- a/pykob/preferencesWindow.py
+++ b/pykob/preferencesWindow.py
@@ -54,7 +54,10 @@ class PreferencesWindow:
         PORT_DEFAULT = 7890
         
         self.CHARACTER_SPACING_OPTIONS = ["None", "Between characters", "Between words"]
-        self.CHARACTER_SPACING_SETTINGS = ['NONE', 'CHARACTER', 'WORD']
+        self.CHARACTER_SPACING_SETTINGS = ['NONE', 'CHAR', 'WORD']
+        self.CHARACTER_SPACING_NONE = 0
+        self.CHARACTER_SPACING_CHARACTER = 1
+        self.CHARACTER_SPACING_WORD = 2
         self.DEFAULT_CHARACTER_SPACING = 2
         
         self.CODE_TYPES = ["American", "International"]
@@ -222,29 +225,40 @@ class PreferencesWindow:
         
         ttk.Label(codeOptions, text="Code speed and Farnsworth spacing:").grid(row=0, column=0, columnspan=2, sticky=tk.W)
         
-        ttk.Label(codeOptions, text="Intial Code Speed:").grid(row=1, column=1, padx=30, sticky=tk.E)
+        ttk.Label(codeOptions, text="Code Speed:").grid(row=1, column=1, padx=30, sticky=tk.E)
       # print("Setting code speed to", config.text_speed)
         self.codeSpeed = tk.DoubleVar(value=config.text_speed)
         ttk.Spinbox(codeOptions, from_=1, to=99,
                     width=4, format="%2.f", justify=tk.RIGHT,
+                    command=self._updateSpacingOptions,
                     textvariable=self.codeSpeed).grid(row=1,
                                                       column=2,
                                                       padx=10, sticky=tk.W)
         
-        ttk.Label(codeOptions, text="Character rate:").grid(row=1, column=3, sticky=tk.E)
+        ttk.Label(codeOptions, text="Dot speed:").grid(row=1, column=3, sticky=tk.E)
         self.characterRate = tk.DoubleVar(value=config.min_char_speed)
-        ttk.Spinbox(codeOptions, from_=1, to=99, width=4, format="%2.f", justify=tk.RIGHT, textvariable=self.characterRate).grid(row=1, column=4, padx=10, sticky=tk.W)
+        ttk.Spinbox(codeOptions, from_=1, to=99, width=4, format="%2.f", justify=tk.RIGHT,
+                    command=self._updateSpacingOptions,
+                    textvariable=self.characterRate).grid(row=1, column=4, padx=10, sticky=tk.W)
         
         # Create three Radiobuttons using one IntVar for the character spacing options
         ttk.Label(codeOptions, text="Character spacing:").grid(row=2, column=0, columnspan=4, sticky=tk.W)
         # Initialize the code spacing option to its default value of 'None':
         self.characterSpacing = tk.IntVar(value=self.DEFAULT_CHARACTER_SPACING)
+        # Preserve the currently configured Farnsworth spacing option to restore
+        # it if the dot speed is toggled to be the same as code speed or not
+        self._original_configured_spacing = self.CHARACTER_SPACING_CHARACTER + 1
+
+        self._spacingRadioButtonWidgets = []
         for spacingRadioButton in range(len(self.CHARACTER_SPACING_OPTIONS)):
-            ttk.Radiobutton(codeOptions, text=self.CHARACTER_SPACING_OPTIONS[spacingRadioButton],
-                            variable=self.characterSpacing,
-                            value=spacingRadioButton + 1).grid(column=1, sticky=tk.W)
+            self._spacingRadioButtonWidgets.append(
+                ttk.Radiobutton(codeOptions, text=self.CHARACTER_SPACING_OPTIONS[spacingRadioButton],
+                                variable=self.characterSpacing,
+                                value=spacingRadioButton + 1))
+            self._spacingRadioButtonWidgets[spacingRadioButton].grid(column=1, sticky=tk.W)
             # If current config matches this radio button, update the selected value
             if config.spacing.name.upper() == self.CHARACTER_SPACING_SETTINGS[spacingRadioButton]:
+                self._original_configured_spacing = spacingRadioButton + 1
                 self.characterSpacing.set(spacingRadioButton + 1)
     
         # Create a pair of Radiobuttons using one IntVar for the code type options
@@ -259,7 +273,9 @@ class PreferencesWindow:
             # If current config matches this radio button, update the selected value
             if config.code_type.name.upper() == self.CODE_TYPE_SETTINGS[codeTypeRadioButton].upper():
                 self.codeType.set(codeTypeRadioButton + 1)
-    
+
+        self._updateSpacingOptions()
+
       # codeOptions.grid(row=2, column=0, columnspan=4, pady=6, sticky=tk.W)
         codeOptions.pack(fill=tk.BOTH)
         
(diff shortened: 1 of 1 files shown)


pwdirks:
This fixes any inconsistencies between selecting spacing and specifying the same or different “dot speed” vs. “code speed”.  There’s one other thing we could do to really make it clear what’s happening: since there’s no way to send with “dot speed” > “code speed”, we COULD make the preferences window automatically bump up “code speed” to keep up with “dot speed” as soon as “dot speed” = “code speed”?  That would have the effect of ensuring code speed is always at least dot speed with dot speed increases ultimately “raising the ceiling” and dragging overall code speed with it.

Thoughts?

73, -Patrick.

leskerr:
>since there’s no way to send with “dot speed” > “code speed”

Why is it not possible to send with 'dot speed' > 'code speed'? That's the normal situation with Farnsworth. For example, you might want to send your text with a code speed of 15 wpm, but have each character sent at dot speed of 18 wpm.

leskerr:
This branch works for me. I might rather have full control and be able to turn off Farnsworth temporarily by selecting 'None' for spacing when 'Code speed' is, say, 15 wpm and 'Dot speed' is 18 wpm, but I can achieve the same result by cranking 'Dot speed' down to 15 and then cranking it back up to 18 (assuming I remember where it used to be). But if there are advantages to disabling 'None' I can definitely live with it.

pwdirks:
Sorry - that’s stated exactly backwards.  You can’t select “dot speed” LESS THAN “code speed”.  So as you dial dot speed BELOW code speed, the UI could automatically drag code speed down with you.

pwdirks:
When you imagine (temporarily) turning Farnsworth OFF, what do you think you have in mind?  "Let me see what this code speed looks like full-out”?  Or "let me see what that speed when sent without continuous looks like”?  Or “Let me dial back the pauses and see how it goes”?

One might lend itself more naturally to adjusting code speed - the other perhaps to “forcing” code speed to match the selected dot speed.  We could permit both but I wonder if people would be surprised if clicking on the “None” radio button jumped the code speed ahead?  They might even miss the effect, and then be surprised it’s not what they set it to?

73, -Patrick.

leskerr:
> Sorry - that’s stated exactly backwards. You can’t select “dot speed” LESS THAN “code speed”. So as you dial dot speed BELOW code speed, the UI could automatically drag code speed down with you.

But I always set my Farnsworth dot speed less than the code speed. My standard settings are code speed = 24 and (minimum) dot speed = 18. As always, code speed takes precedence and the dots are actually sent at 24 wpm. If I occasionally want to send much slower, at 15 wpm say, then I change my WPM setting on the main KOB window to 15. Farnsworth kicks in and the characters are sent at 18 wpm. When I want to go back to my normal speed, I change the WPM setting to 24. I never touch the Morse Code tab after first setting it up.

leskerr:
> When you imagine (temporarily) turning Farnsworth OFF, what do you think you have in mind?

As always, the code speed setting takes precedence. So when I turn Farnsworth off, the dot speed setting is ignored, the code is sent at the specified code speed without any extra space between characters.

pwdirks:
… but that results in seemingly impossible settings (code speed = 24/dot speed = 18 or code speed = 15/dot speed=18/Farnsworth=None).  YOU know the code speed settings take precendence but I’m not sure that’s a good UI - the settings don’t show how they’re actually being interpreted.

I’d rather if you set a dot speed greater than the actual code speed, you see there’s some spacing inserted somewhere.  Conversely, when the speeds are the same, you should see that spacing=None.  When you decide you want to send more slowly, I think it’s natural to start dialing down code speed; when it goes below dot speed, you should see spacing come alive and show you where it’s being inserted.

I guess the main oddity is you want to have a Farnsworth speed specified BELOW your code speed, even though that can’t really be done, just to “peg” the dot speed for when you control the code speed?

If we make decreasing dot speed drag down code speed with it, and increasing code speed drag dot speed UP with it to ensure “dot speed” setting is always >= “code speed” setting, I guess you’d end having to dial down both the code speed AND the dot speed some, from both =24 to code speed=15, dot speed=18.

Let me try tying those speeds together to enforce their relative settings and let’s experiment with dialing it around?  Let’s see how your brain reacts :-)

73, -Patrick.

leskerr (approved):
I've tried it and I can do everything I want to do, even though it's clumsy

pwdirks:
Hi Les,

Thanks for reviewing the change.  I’d like to see if we can do better… what do you find clumsy?  Not manipulating Farnsworth spacing directly with speed settings possibly inconsistent?  Is that something you think your brain might adjust to?  Should we try something different (perhaps as a separate change)?

Did you try the latest commit, BTW, which drags “dot rate” up as code speed goes above “dot speed” and drags code speed down as dot speed is slowed BELOW code speed?

73, -Patrick.

leskerr:
Hi Patrick,

Yes, I pulled the latest commit (with the dragging) and tried it. And yes,
my brain has adjusted to it.

There may still be an untidiness. If I change my code speed from 25 to 30
on the main KOB window and then open the Morse Code tab. It shows code
speed 30 and dot speed 25. I thought you were going to have the dot speed
track the code speed when the code speed increases.

~Les

pwdirks:
Hi Les,

The immediate update of the running MKOB settings based on config changes made it being tracked as a separate issue.  Ed has made some progress on that separately - I’m waiting for him to submit that but this change should have no impact on that (except perhaps for variable name changes - I did add an underscore to internal preferences variables but we’ll sort that out in the merge if it’s a factor).

I’ll see if Ed has any comments; otherwise I’ll merge the change sometime tonight or tomorrow.

73, -Patrick.

AESilky:
I am working on the updates tonight and hope to finish them by EOD tomorrow.

I will say that the method of setting is somewhat backward to what I would do (I would set Text WPM and Spacing, if Spacing is NONE, then Dot speed doesn't matter. Otherwise, set Dot speed), but I'm a novice, so I defer to the more experienced telegraphers in the group. I guess that one wrinkle is that the main UI only has the Text WPM value. So you can adjust it there without being able to make any changes to Spacing or Dot Speed. So what happens as you increase/decrease the Text WPM value without having the ability to change either the Spacing or Dot Speed values from what they had been set to?

pwdirks:
Hi Ed,

Actually I was just thinking tonight how there might be an alternative approach as well.

Let’s call the current approach “speed first” - you configure the code speed and the dot speed and the UI makes sure whatever you specify is sensible and physically possible.  Farnsworth spacing is disabled if not applicable; users can opt for “character” or “word” spacing if Farnsworth is needed.

The alternative would be, let’s call it “spacing first” - the Farnsworth spacing options are always available; if set to “None” then the dot speed either follows code speed 1:1 or is disabled altogether (probably more sensible), leaving just the code speed to tweak.  If you select “Character” or “Word spacing” the “dot speed” control comes alive as a separate control (still tied to code speed based on physical limits to make sure "dot speed” >= "code speed” at all times) but all Farnsworth controls are always available.  The UI still ensures only physically realizable settings can be entered.

I think it makes sense at any time to select “Character” or “Word” spacing and enable both controls.  If you have two different speeds up with, say, “Character” spacing selected and you change the Farnsworth spacing to “None”, presumably code speed and dot speed become the same.  Which speed should be used, though?  Should code speed increase to dot speed?  Or dot speed slow down to code speed?

If you permit me to bias the discussion: since the dot speed control is about to be disabled, I’d vote to leave that one unchanged and increase code speed to match dot speed - now that that control is “live” you can change the setting if you don’t like what it “jumped up” to?  I’d think it’d be awkward to have the dot speed suddenly change but then be disabled and you have to think to realize that you can get whatever you were hoping for by tweaking the OTHER speed setting (code speed).

Thoughts?  Want me to whip up an experiment on another branch?

73, -Patrick.

leskerr:
I'm going to step back from this discussion and not try to guess what's best for most users. My comments will be limited to what works for me and what doesn't.

The fact of the matter is, I NEVER use Farnsworth (except for my feeds, which don't use the MKOB client anyway). I want to be able to turn Farnsworth off, and have it stay off no matter what I do with the WPM control on the main KOB window.

For me, the problem with the current branch is if I'm running with WPM at 24 (and spacing 'None') and I lower WPM to 22, 'Dot speed' stays stuck at 24 and spacing becomes 'Between characters'. This is unacceptable.

~Les

AESilky:
@pwdirks you say:
> The alternative would be, let’s call it “spacing first” - the Farnsworth spacing options are always available; if set to “None” then the dot speed either follows code speed 1:1 or is disabled altogether (probably more sensible), leaving just the code speed to tweak.  If you select “Character” or “Word spacing” the “dot speed” control comes alive as a separate control (still tied to code speed based on physical limits to make sure "dot speed” >= "code speed” at all times) but all Farnsworth controls are always available.  The UI still ensures only physically realizable settings can be entered.

That is what I intended to suggest in my earlier comment in our email discussion (2/3/2021):
> It might be a difference between an experienced telegrapher and a novice - but I think it would be more natural to me to allow specifying the 'Text' WPM and then have a 'Farnsworth' section. The `spacing` value selection would have a label of 'Farnsworth' with possible values of 'None'/'Between Characters'/'Between Words' (representing `SPACING`='NONE'/'CHAR'/'WORD'). The 'Dot Speed' selection would be within the Farnsworth section, but would only be enabled if the selected value of 'Farnsworth' isn't 'NONE'. If the 'Dot Speed' selection is enabled the value should initially be populated with the currently configured dot speed if it is greater than the entered text speed or with the entered text speed (not the currently configured text speed). Any dot speed value entered should be checked to validate that the entered value is greater than (or equal to) the entered 'Text Speed' value. Otherwise, that value shouldn't be allowed.

I like that approach because to me it is more intuitive for a new user of the application and I think it will also solve @leskerr 's situation where he sets 'SPACING=NONE' and controls the speed entirely with the 'TEXT WPM' value on the main GUI.

Also, this says it fixes #261, but that is the 'Local sounder' issue. I believe you mean #260.

pwdirks:
I don't want to move this discussion to a separate pull request but I've created a separate branch to try the alternative.  Check out branch "pwd-260-preferences-with-new-Farnsworth-model" sometime and let me know how you find that model?

The only thing I had to make a call on was whether to keep dot speed >= code speed by bumping it up as code speed is increased as the previous change did.  I decided against that in favor of a one-time check when Farnsworth spacing is selected: at that point code speed will drop to match dot speed (previous value unchanged) as the separate dot speed control is enabled.

Please let me know what you think!

AESilky:
I changed the subject to refer to the correct issue #260.

leskerr:
The pwd-260-preferences-with-new-Farnsworth-model branch works for me.

~Les

pwdirks:
Great!  @leskerr if this version works for you let me go ahead and merge pwd-260-preferences-with-new-Farnsworth-model in instead of this.  I'll should be able to do the merge later this evening.

pwdirks:
Fixed by merging "Fix #260 by renaming speed controls and tying availability of separate 'dot speed' to Farnsworth spacing setting" branch instead.

PR #264, opened 2021-02-08
Handle '-32767' value is 'speedfactor' check.

### LINKED ISSUE #263: PyKOB Play (recorder) --playback (speed) option doesn't seem to work
@leskerr reported:
> I'm trying to play back a recorded session at half speed with this command:
>
> py play.py --speedfactor 50 Session-20210206-0033.json
>
> It still plays at full speed. Am I doing something wrong? I think it used to work for me.

This is the correct use of the command line option and should cause the recording to be played back at half speed.

AESilky:
Added a check for the `-32767` 'special' value. This seems to correctly adjust the playback, but I would like an experienced ear to validate the result.

Closes #263

### DIFF
--- a/pykob/recorder.py
+++ b/pykob/recorder.py
@@ -576,9 +576,7 @@ class Recorder:
                                 time.sleep(pause)
                         if not self.__speed_factor == 100:
                             sf = 1.0 / (self.__speed_factor / 100.0)
-                            for c in code:
-                                if (c < 0 or c > 2) and c != -32767:
-                                    c = round(sf * c)
+                            code[:] = [round(sf * c) if (c < 0 or c > 2) and c != -32767 else c for c in code]
                         self.wire = wire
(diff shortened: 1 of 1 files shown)


AESilky:
If this change doesn't result in the %50 slowdown, it's possible that the KOB playback doesn't 'like' the (much) slower speeds requested by the values.

If that is the case @AESilky and @leskerr will probably need to troubleshoot/modify together.

AESilky:
I will say that it 'sounds slower to me', but I will add that I don't yet have 'the ear' to really discern the difference. I will need @leskerr to listen and comment on the result.

If this doesn't result in the expected 1/2 speed playback, I will probably need @leskerr 's help to step through it and figure out where the 50% isn't being applied correctly.

leskerr (approved):
It works fine now. Thanks!

PR #268, opened 2022-05-02
Add GPIO (Raspberry Pi) support.

AESilky:
First push for testing on different devices.


AESilky:
Tested with Raspberry Pi KOB HAT interface.

Issue #273, opened 2022-11-08
Sounder not working with loop interface and serial


AESilky:
When using my KOB (loop interface) though a serial cable, I found that the sounder wasn't working, even though it was enabled. This probably got broken while working on the GPIO interface support.

PR #274, opened 2022-11-09
Fix sounding with loop interface. Support disabling local sounder.

AESilky:
This was broken as part of the changes made to support the GPIO interface. This fixes the probably and also the feature to disable the local sounder.

#273 and #261


AESilky:
This closes #273
This closes #261

PR #276, opened 2022-12-14
Improve code decoding when large Farnsworth is involved (not a comple…

AESilky:
…te fix).

Change `Receive` to take `cwpm` in addition to `wpm` and use the larger of the two to decode with.

### DIFF
--- a/Receive.py
+++ b/Receive.py
@@ -32,9 +32,6 @@ To maintain backward compatibility it also allows a positional command
 line parameter:
     1. KOB wire no.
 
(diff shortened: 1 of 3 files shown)


AESilky:
I tested this a number of times and it is significantly better, so I'm going to go ahead and merge this. I'm trying to minimize the changes to *master* until the (large) 'Config' change is complete, so I'll hold off on trying to resolve the occasional 'ME' rather than 'M O' decoding until later.

Issue #279, opened 2023-11-19
Receive.py 'unexpected keyword' error


AESilky:
The following occurs when running Receive.py to receive from a wire:

Traceback (most recent call last):
  File "/Users/esilky/proj/morse/pkWIP/PyKOB/Receive.py", line 109, in <module>
    myReader = morse.Reader(wpm=text_speed, code_type=code_type, callback=readerCallback)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: Reader.__init__() got an unexpected keyword argument 'code_type'

PR #280, opened 2023-11-20
Correct values passed to morse.Reader.

AESilky:
A named argument name was incorrect, and the value being passed was a string rather than an ENUM. Fixed both issues.


AESilky:
Closes #279

Issue #286, opened 2023-12-15
Key timing graph - graph bar length too long for spaces


AESilky:
On Windows, the space graph lines are longer than the mark graph lines for an equivalent duration. On Mac and Linux they are correct.

AESilky:
Even though a fixed pitch font is being used, on Windows the 'solid block' character was being rendered narrower than the other block characters (specifically in this case, the 'line'). Switching to the 'dark shaded block' (rather than the 'solid block') fixed the rendering size difference. I would rather have the solid block, but the dark shade is acceptable.

PR #287, opened 2023-12-15
Key timing graph - use dark shade rather than solid block for 'mark' …

AESilky:
…bars (Windows block is smaller for some reason).


AESilky:
Closes #286

PR #291, opened 2023-12-28
Move local sounder drive from key into the key handling loop to avoid…

AESilky:
… delay. Fix start-up issue with mkob class restructure and auto-connect.

### DIFF
--- a/Configure.py
+++ b/Configure.py
@@ -100,6 +100,7 @@ def main(argv):
             config.interface_type_override, \
             config.invert_key_input_override, \
             config.min_char_speed_override, \
+            config.local_override, \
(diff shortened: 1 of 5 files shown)


AESilky:
My separate key & sounder interface is packed away, so I could only test this with key & virtual sounder (computer audio), but that works well, so I think the actual sounder will as well.

The other fix is for a bug that was introduced with the refactor/reorganization of all the MKOB modules into true classes. The problem occurs during start-up when the 'auto-connect' option is enabled.

leskerr:
That fixed the problem. Thanks, Ed!

~Les

PR #401, opened 2025-09-06
KOB: Add option to extend the Key Closer open check time.

AESilky:
The Alberta-MTC installation has seen cases where a station will open the wire when nobody is actually operating the station. It is thought that this is being caused by noise on the physical loop circuit causing MRT/MKOB to think that the key is being opened.

Added this option to require that the line/circuit be continuously open for 8/10 second before considering the circuit open.


AESilky:
Hopefully this will solve the problem with Midnapore opening and sending 'E' when nobody is there.

jchausler:
Hi Ed,

I'm assuming that by 8/10 you mean 0.8 of a second.  When I first read it I thought you meant 8 to 10 seconds, way too long. 

I don't  know how long their loop circuit is at the the Midnapore station (to my knowledge the only place this is happening).  The longer it is, the greater chance of the loop wires acting as an antenna.  If noise it the problem, and that would need to be determined, putting a capacitor across the loop at the interface might trap it. One popular WU spark killer circuit back in the daze put a capacitor across the sounder (and then a 20 Ohm resistor in series with the sounder).  As to killing sparks, not quite as effective as a TVS diode but I used it early on and it did noticeably help.  That said the only experiences with loop noise I ever had would either lock up the USB-to-Serial interface or the USB interface in the computer itself usually requiring a reboot, but never causing spurious short time circuit openings.  If someone there has access to an oscilloscope, that would be a good tool to use to look for a noise problem.

But the engineer in me says that it is much better practice to find the source of a problem and fix it there then trying to bury it with some kind of work-around (in software I've been there, done that, but not always completely successfully and it never was good practice.  That said, desperation can make for strange bedfellows :-)

73, Chris
-----Original Message-----
From: MorseKOB/PyKOB ***@***.***>
Sent: Sep 6, 2025 5:40 PM
To: MorseKOB/PyKOB ***@***.***>
Cc: Subscribed ***@***.***>
Subject: [MorseKOB/PyKOB] KOB: Add option to extend the Key Closer open check time. (PR #401)

The Alberta-MTC installation has seen cases where a station will open the wire when nobody is actually operating the station. It is thought that this is being caused by noise on the physical loop circuit causing MRT/MKOB to think that the key is being opened.
Added this option to require that the line/circuit be continuously open for 8/10 second before considering the circuit open.
You can view, comment on, or merge this pull request online at:  https://github.com/MorseKOB/PyKOB/pull/401
Commit Summary  *  430c0a0 (https://github.com/MorseKOB/PyKOB/pull/401/commits/430c0a02c34c1f5fabc3e2eab2f35a0619077ff2) KOB: Add option to extend the Key Closer open check time.

File Changes(13 files (https://github.com/MorseKOB/PyKOB/pull/401/files))
  *  M Documentation/MKOB/User-Manual-MKOB4.adoc (https://github.com/MorseKOB/PyKOB/pull/401/files#diff-f3c6e08ea02e2ef068ddf251750a79dbdbe2991f609c99d8f4a10a8717f431c3) (4)
  *  M Documentation/MKOB/User-Manual-MKOB4.pdf (https://github.com/MorseKOB/PyKOB/pull/401/files#diff-5c77e3e13507fb593635e3d6ebb432d4bc1c59a53d003e294ca9f6f4c5d380f9) (0)
  *  M Documentation/MKOB/images/preferences-advanced.png (https://github.com/MorseKOB/PyKOB/pull/401/files#diff-5a2697f1f3f90be8c5567cfc44b2fc788e6ec89a3574ba003731aeb838407122) (0)
  *  M mkobversion.txt (https://github.com/MorseKOB/PyKOB/pull/401/files#diff-55df4cfc2fe7c5da0ed9d3cdbd3fbd34c38f975525d418033b68d1f5dcddb1a5) (2)
  *  M msuiteversion.txt (https://github.com/MorseKOB/PyKOB/pull/401/files#diff-93cb76dd16e60159f7a69fbc8281fe70d4e35f1ef59d2d7979abe0e195fc8c40) (2)
  *  M src.py/Configure.py (https://github.com/MorseKOB/PyKOB/pull/401/files#diff-f0bb176d15ca337047c3afb5c6986e7bb51f673136bac7a9367e8d690dbfcdff) (1)
  *  M src.py/MKOB.py (https://github.com/MorseKOB/PyKOB/pull/401/files#diff-0905a8604fa09d97a5fd58ebc2ccbb2a7b0f70c6e06b4aab5b4c74d670147dd7) (2)
  *  M src.py/mkobmain.py (https://github.com/MorseKOB/PyKOB/pull/401/files#diff-b5ea6506aed80ab99041c1d3c032571f4c21b8dc81edb2cbe521366c0be25454) (5)
  *  M src.py/pykob/__init__.py (https://github.com/MorseKOB/PyKOB/pull/401/files#diff-671f571c62d1274e281aa989d38c7f21e7354a73f03c0e06cb32937ddda121be) (2)
  *  M src.py/pykob/config.py (https://github.com/MorseKOB/PyKOB/pull/401/files#diff-4da8b618c76213ea00036214b03cfb95535f82d64fa09c7e65f81d58872fe4c1) (40)
  *  M src.py/pykob/config2.py (https://github.com/MorseKOB/PyKOB/pull/401/files#diff-c553bde6a525f421dcdef63cddc314784243a1e74980bb386c668d1ac3bfa57b) (46)
  *  M src.py/pykob/kob.py (https://github.com/MorseKOB/PyKOB/pull/401/files#diff-60cd971819804c6d2747be4f831039ce79455ff64612fd5712c1d4969b33b5f8) (54)
  *  M src.py/pykob/preferencesWindow.py (https://github.com/MorseKOB/PyKOB/pull/401/files#diff-45f8dcf89fa955a4757dd8ebadf221e66b9b39cd898fda0a44a1f37e07bd9f9e) (6)

Patch Links:  *  https://github.com/MorseKOB/PyKOB/pull/401.patch
  *  https://github.com/MorseKOB/PyKOB/pull/401.diff

&mdash;

jchausler:
Hi Ed,

I wasn't suggesting that you go to Calgary to fix this (although Calgary is a nice place to visit, I spent most of the first half on 1998 there installing a control system for Canadian Pacific :-) but rather that Kevin or someone else with the appropriate skills more local to the site check out the hardware (of course us software guys always think it's the hardware ;-)  But I'd be suspicious of a loose connection somewhere as I have been on their wire when this occurs although the regularity of it makes me wonder.  But I have had such problems with intermittent connections.  The 1908 Vibroplex at the AWA museum's circuit closer occasionally just momentarily opens for reasons I have been unable to discern but usually only a single occasional event and after all that bug is old and beat.  I think it is due to corrosion and wiggling it a bit seems to correct the problem (until it happens again).

That said, after I wrote the below it suddenly dawned on me that when I am sending a string of dots, the open time between the closed time for a string of dots runs on the order of only 0.05 seconds.  I'm assuming you are differentiating between the first open after an extended close time and normal sending activity as otherwise 0.8 seconds would be way too long to ignore.

The "real world" can be a nasty place into which to install technology.  Bad solder joints in the interface could also be a problem with any jostling.  In a control system I was involved with years ago, we had an intermittent input toggling but then it would stop.  It was due to a slight vibration shaking the racks in which the equipment was installed causing an intermittent connection and would only occur when a cooling fan came on.  In another installation (for CP in St. Paul :-) the system was on the second floor over a large air compressor and when the compressor came on, it would shake everything.  It had a habit of destroying hard drives due to causing head crashes and it quickly found loose wires as well (but it was kind of nice to sit in a chair there when this occurred as it was like getting a massage in one of those vibrating chairs :-)  I can even describe another situation where the vibration was so bad that on printed circuit cards component leads would eventuall break on larger components (capacitors) due to the repeated accelerations of the body of the component.

If I was them and I didn't have any diagnostic tools (but what does a digital VOM cost?) I would start by swapping out everything in the Midnapore installation including the wiring and then, piece-by-piece, putting it back until the failure returns.  This would of course take some significant waiting time.  Anyway, I've personally been there (many times), done that, and don't want to have to ever do it again ;-)

73, Chris

-----Original Message-----

AESilky:
Ha ha Chris,
Since I've worked as both a hardware and a software engineer, I end up
pointing a finger at myself! ;-)

Yes, as I pointed out in my comment, the additional time is only added when
going from a 'key/circuit closed' state to a 'key/circuit open' state. Once
the circuit is open, just the standard debounce of 0.008 is used.

AESilky:
STATUS UPDATE (2025/9/10)

Calgary-MTC has reported that, since installing the update and enabling the option, no spurious circuit opening is showing up from the Midnapore station and that someone at Midnapore was able to use the key to send without any problem.

PR #407, opened 2026-06-16
Loop power cut issues

### LINKED ISSUE #404: MRT: Using the option to add DT to Sender output not working when using a Selector Switch
With MRT (and MKOB) the command line option, `--senderdt` causes the current date and time to be added to the current sender station name/id when the sender changes. When using MRT with a Selector Switch this command line option doesn't work.

As a work-around, the `--senderdt` can be added into each of the selector switch configurations, but if it is on the main command line, it should be inherited by the selector switch configurations.

AESilky:
Fixes for not being able to break-in to an open wire from MRT and MKOB.
Various small fixes that were staged.


AESilky:
closes issue #404

